PR #1, opened 2021-04-01
Launch page

Peartes:
[x] Create a layout for  the launch page
[x] Style the launch page


src/launch/launch.scss
```diff
@@ -1,3 +1,110 @@
 @import "styles/colors.scss";
+@import "styles/styles.scss";
```
dkent600, line 2: you can delete this line, styles is included by app.tx, unless, i guess, it is needed by the @extend things

src/launch/launch.scss
```diff
@@ -1,3 +1,110 @@
...
 .page.launch {
+  display: grid;
+  justify-content: center;
+  grid-gap: 30px;
+  .banner {
+    .banner1 {
+      // Make like a heading
+      @extend .heading;
+      font-size: 4vw;
```
dkent600, line 11: hmmm, haven't used vw for a font size before.  but it could make for easy responsiveness.  gotta see that

src/launch/launch.scss
```diff
@@ -1,3 +1,110 @@
...
+      display: grid;
+      justify-content: center;
+    }
+    // TODO: Customize for mobile
+    .banner2 {
+      @extend .subtext;
+      // TODO: Get font from figma
+      font-family: "Lucida Sans", "Lucida Sans Regular", "Lucida Grande",
```
dkent600, line 19: limit yourself to the two fonts we have been using: Aeonik and Cera Pro

src/launch/launch.scss
```diff
@@ -1,3 +1,110 @@
...
+      justify-content: center;
+      text-align: center;
+    }
+  }
+  // Let's define layouts for the launch divs
+  .launch-divs {
+    display: flex;
+    flex-flow: row wrap;
+    flex: 1;
+    width: 100vw;
```
dkent600, line 31: i'm skeptical about the 100vw here

src/launch/launch.scss
```diff
@@ -1,3 +1,110 @@
...
+    align-items: center;
+    justify-content: space-evenly;
+  }
+  // ** For styling the container holding buttons
+  .div-style {
```
dkent600, line 36: I would not use "div-" in the style names, nor in the name of the custom element you created.  better to leave the implementation out of the names, cause it could change

src/launch/launch.scss
```diff
@@ -1,3 +1,110 @@
...
+      background-color: #2e5aff;
+      font-size: 0.7em;
+      border-radius: 20px;
+      // TODO: Colors need to be updated in global style
+      &:hover {
+        cursor: pointer;
+      }
+    }
+    a {
+      font-size: 0.8em;
```
dkent600, line 86: I've been using px for font sizes lately, but wanting to switch to em, so I will start trying to do that.

dkent600 (changes requested):

src/launch/launch.scss
Peartes: yes, responsive font sizes (thinking about mobile)

src/launch/launch.scss
Peartes: okay

src/launch/launch.scss
Peartes: Yes, the grid layout uses this width min-content that just packs things together. Check it out

src/launch/launch.scss
Peartes: Aha, I wanted to change this anyways

src/launch/launch.scss
Peartes: Great! i guess i'll also do more em (when needed also)

src/launch/launch.scss
Peartes: It is needed. i just removed it and realised it was

Peartes (commented):
Made the changes to the fonts

src/launch/launch.scss
dkent600: vw is usually defined to include the scrollbar.  is that not the case here?

src/launch/launch.html
```diff
@@ -1,5 +1,31 @@
...
   <div class="page animated-page au-animate launch">
-    Launch
+    <div class="banner">
+      <div class="banner1">
+        <span> Apply to host </span>
+        <span style="padding-left: 1vw"> your Launch </span>
+      </div>
+      <div class="banner2">
+        <span>
+          <p>Prime Launch empowers and creators to host and fair and</p>
```
dkent600, line 11: I believe it is best not to use 'p' because it may be implemented differently o different browsers.  better to stick with div and a class

dkent600 (changes requested):

dkent600 (approved):

PR #2, opened 2021-04-03
Added the select a package page

Peartes:
[x] Added UI components


src/app.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable linebreak-style */
```
dkent600, line 1: why is this added?

src/selectPackage/selectPackage.ts
```diff
@@ -1,3 +1,26 @@
+/* eslint-disable linebreak-style */
 import "./selectPackage.scss";
+
 export class SelectPackage {
+  // Declare state variables
+  benefits: Array<string>;
+  discounts: { [plan:string]:{discount: number, benefits: Array<string>}};
```
dkent600, line 7: Why this complication in the ts?  Can this just be hardcoded in the HTML?  Same with benefits.

src/selectPackage/selectPackage.scss
```diff
@@ -1,3 +1,97 @@
...
+    //   font-size: 0.7em;
+    //   border-radius: 20px;
+    //   // TODO: Colors need to be updated in global style
+    //   &:hover {
+    //     cursor: pointer;
+    //   }
+    // }
+    a {
+      font-size: 0.8em;
+      text-decoration: underline;
```
dkent600, line 76: underline should be the default for all anchor.

src/selectPackage/selectPackage.scss
```diff
@@ -1,3 +1,97 @@
...
+      // TODO: Wrong color
+      text-decoration-color: #2e5aff;
+      text-decoration-thickness: 0.15em;
```
dkent600, line 79: hmmm, is this really needed?  it is very unusual.

src/selectPackage/selectPackage.scss
```diff
@@ -1,3 +1,97 @@
...
+          color: firebrick;
+          font-size: 1.3em;
+          padding: 0.4em;
+        }
+      }
+    }
+    .buttonContainer a {
+      text-decoration: none;
+      // TODO: Remove this
+      color: white;
```
dkent600, line 59: I think we should always use color variables from _colors.scss.

src/selectPackage/selectPackage.scss
```diff
@@ -1,3 +1,97 @@
 @import "styles/colors.scss";
+@import "styles/styles";
```
dkent600, line 2: this should not need to be imported

dkent600 (changes requested):

src/selectPackage/selectPackage.ts
dkent600: aside from the complication of add'l code, realize that aurelia will create observables on these unless you specify them to be "one-time".  but then again, it begs the question why have them at all?

src/app.ts
Peartes: 🤦🏽‍♂️  it's because of difference in our linebreak char code. My linter always marks the line-break as an error. That is just to make it stop

dkent600: go ahead and delete this line and pay no attention to the lint error for now. I will convert everything to LF and the error will go away for you.

dkent600 (changes requested):

src/app.ts
Peartes: 👌🏽

src/selectPackage/packageContainer.html
```diff
@@ -0,0 +1,29 @@
...
+        <div class="benefit"><span>&#10026;</span> Hosted</div>
+        <div class="benefit"><span>&#10026;</span> Marketing</div>
+        <div class="benefit">
+          <span>&#10026;</span> 25% discount on services
+        </div>
+      </div>
+    </div>
+    <div class="packageFooter">
+      <div class="heading heading2 discount">
+        ${discount > 0 ? discount + "% "+ "FEE" : "FREE"}
```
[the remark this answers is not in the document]
Peartes: 👌🏽

dkent600 (approved):

PR #3, opened 2021-04-13
Newseed Stage 1

Peartes:
[x] Stage 1 layout
[x] Basic validation 
[o] Still need to include functionality for adding more rows


src/newSeed/baseStage.scss
```diff
@@ -0,0 +1,223 @@
+@import "styles/colors";
+// This styles are common to all the stages components
+
+.page.stageContainer {
+  &.stage1 {
+    .infoContainer {
+      background-color: $color-black-2;
+      padding: 2em;
+      border-radius: 15px;
```
dkent600, line 9: em?

src/newSeed/stage1.ts
```diff
@@ -1,5 +1,33 @@
...
+    } else if (!Utils.isValidUrl(this.seedConfig?.general?.projectWebsite)) {
+      message = "Please enter a valid website url";
+    }
+    else if (!this.seedConfig?.general?.category) {
+      message = "Enter a category";
+    } else if (!this.seedConfig?.general?.whitepaper) {
+      message = "We need your whitepaper address";
+    } else if (!Utils.isValidUrl(this.seedConfig?.general?.whitepaper)) {
+      message = "Please enter a valid website url";
+    } else if (!this.seedConfig?.general?.github) {
```
dkent600, line 22: delete this check and replace by padding `false` as the 2nd param to isValidUrl

dkent600 (changes requested):
Can you please remove the changes to stages > 1 from this PR?  I would prefer to review them one at a time.

src/newSeed/baseStage.scss
Peartes: Yes 😂 to resize with font size

src/newSeed/stage1.ts
Peartes: Delete the check ? or just add a false as a second param ?

src/newSeed/baseStage.scss
Peartes: It allows for better responsiveness and i investigated that

Peartes:
I made the changes so you can review stage 1 independently @dkent600

src/newSeed/baseStage.scss
dkent600: the comment is on a line that is using px

src/newSeed/stage1.ts
dkent600: Delete the check.  note what false does

dkent600: don't forget to update the error message

src/newSeed/baseStage.scss
Peartes: Aha, for border-radius

PR #9, opened 2021-04-19
Newseed

Peartes:
Make the dynamic social link component working
Themed the newseed page

PR #10, opened 2021-04-21
Newseed: Adjustement to seed2 PR

Peartes:


src/footer/footer.html
```diff
@@ -29,33 +29,33 @@
...
-          <a href="#" target="_blank" rel="noopener noreferrer"
-            >Upcoming Launches</a
-          >
-          <a href="#" target="_blank" rel="noopener noreferrer"
-            >Host a Launch</a
-          >
-          <a href="#" target="_blank" rel="noopener noreferrer"
-            >Prime Launch Docs</a
-          >
+          <a href="#" target="_blank" rel="noopener noreferrer">
```
dkent600, line 32: really these anchors should be on one line.  that is how they were.  as you have it, one is introducing whitespace, with potential consequences (that I don't know how to fully predict).  further I think, within reason, one-liners are cleaner in appearance.  Same for all these similar changes across this file.

src/newSeed/stage1.ts
```diff
@@ -1,21 +1,17 @@
-import { RouteConfig } from "aurelia-router";
 import { BaseStage } from "./baseStage";
 // Let's import the utils helper
 import {Utils} from "../services/utils";
 export class Stage1 extends BaseStage {
-  activate(_params: unknown, routeConfig: RouteConfig): void {
```
dkent600, line 6: good to do this, needs to happen in stage2.ts as well.

dkent600 (changes requested):
Still need:

https://github.com/PrimeDAO/prime-launch-dapp/pull/9#discussion_r616865328

src/footer/footer.html
Peartes: Prettier. I'll increase print width

.vscode/settings.json
```diff
@@ -6,5 +6,7 @@
   "[typescript]": {
     "editor.formatOnSave": false
   },
-  "files.eol": "\n"
+  "files.eol": "\n",
+  "prettier.printWidth": 180,
```
dkent600, line 10: I don't think you should use prettier, and you shouldn't set a rule to limit line lengths.  Much better to just rely on vscode and (es/te)lint.

It is your choice...but I really don't want to deal in PRs with prettier reformating...

dkent600 (changes requested):

dkent600 (approved):

PR #18, opened 2021-05-16
Make seed registration page QA ready

Peartes:
In steps to making the seed registration pages ready for QA, the following changes were made and bugs fixed
    -> Verify validation across all stages: Stages validation confirmed to be working
    
    -> Remove stage 6 form view: Stage 6 should not be visually accessible using the sidebar nav, remove it from the UI and only display it after validation of stages is complete
    
    -> Clean up the launch page: The launch page is the page leading to the seed registration pages. Some UI issue was fixed on this page, mainly the banner was clipping and so it was padded to remove the clip on the text
    
    -> Deal with the navigation bug: The sidebar nav is great :) but introduces a bug into the system where a user could provide valid inputs at first and validate that stage by clicking on the proceed button but then after proceeding goes back to the stage using the nav and changes some input to something else not conforming to set standards and moves to other stages using the nav. in this situation, validation has already been done on this stage and so the stage still remains validated which introduces the bug. The fix was to revalidate the stage upon detaching it from the view and invalidating (no better word known) if any of the changed inputs do not conform to standard
    
    -> Deal with autocomplete style change: The webkit-autocomplete is a great feature really but the background overlay on inputs is a drag. Using the -WebKit selector and some nifty CSS tricks (thanks to stack-overflow :) ) the overlay problem was fixed


src/newSeed/baseStage.scss
```diff
@@ -60,9 +57,9 @@
   label {
     display: block;
     font-size: 1.1em;
+    font-weight: bold;
     line-height: 2em;
-    color: $White;
-    @include standardTextGradient;
+    color: $Secondary02;
```
dkent600, line 62: does this enable is to just include colors.scss instead of styles.scss?

src/newSeed/newSeed.ts
```diff
@@ -35,7 +35,7 @@ export class NewSeed {
       { route: ["stage3"], nav: true, moduleId: PLATFORM.moduleName("./stage3"), name: "stage3", title: "Token Details", settings: { seedConfig: this.seedConfig, stageState: this.stageState, stageNumber: 3, maxStage: 6 } },
       { route: ["stage4"], nav: true, moduleId: PLATFORM.moduleName("./stage4"), name: "stage4", title: "SEED Details", settings: { seedConfig: this.seedConfig, stageState: this.stageState, stageNumber: 4, maxStage: 6 } },
       { route: ["stage5"], nav: true, moduleId: PLATFORM.moduleName("./stage5"), name: "stage5", title: "Contact Details", settings: { seedConfig: this.seedConfig, stageState: this.stageState, stageNumber: 5, maxStage: 6 } },
-      { route: ["stage6"], nav: true, moduleId: PLATFORM.moduleName("./stage6"), name: "stage6", title: "Thank You!", settings: { seedConfig: this.seedConfig, stageState: this.stageState, stageNumber: 6, maxStage: 6 } },
+      // { route: ["stage6"], nav: true, moduleId: PLATFORM.moduleName("./stage6"), name: "stage6", title: "Thank You!", settings: { seedConfig: this.seedConfig, stageState: this.stageState, stageNumber: 6, maxStage: 6 } },
```
dkent600, line 38: I think the proper way to do this is set `nav: false`

src/services/PinataIpfsClient.ts
```diff
@@ -32,7 +32,7 @@ export class PinataIpfsClient implements IIpfsClient {
 
   public async addAndPinData(data: string, name?: string): Promise<Hash> {
     const params = new FormData();
-    params.append("file", Buffer.from(data));
+    params.append("file", Buffer.from(data).toString());
```
dkent600, line 35: Why?  I don't think this is correct.

src/newSeed/stage1.ts
```diff
@@ -13,6 +14,25 @@ export class Stage1 extends BaseStage {
...
+      this.validationError(message);
+      this.stageState[this.stageNumber].verified = false;
+    } else {
+      // For stage 1 write a verified true to stage 1
+      this.stageState[this.stageNumber].verified = true;
+      this.next();
+    }
+  }
+
+  detached(): void {
```
dkent600, line 28: All these `detached` are identical.  they should in the base class.

dkent600 (changes requested):

src/newSeed/baseStage.scss
Peartes: I don't understand but the labels colors were not gradients but just block plain colors hence I removed the text gradient mixin

src/services/PinataIpfsClient.ts
Peartes: Sorry about this, this came from a new commit. The issue here was that the app refused to compile because of some error from that function which I fixed quickly with that just to get the app up and running. I had excluded this file but forgot to exclude it in subsequent commits.

I will revert this back

src/newSeed/stage1.ts
Peartes: I thought of it but they call methods implemented in stage classes

src/newSeed/baseStage.scss
dkent600: can you please try importing colors instead styles at the top of this file, see if it compiles?

src/newSeed/stage1.ts
dkent600: What methods?

src/newSeed/stage6.ts
```diff
@@ -4,4 +4,8 @@ export class Stage6 extends BaseStage {
   close(): void {
     this.router.navigate("/");
   }
+
+  validateInputs(): string {
+    return "";
```
dkent600, line 9: return null here

dkent600 (changes requested):

src/newSeed/baseStage.scss
Peartes: it does compile

dkent600 (approved):

PR #27, opened 2021-05-25
Time input on stage 4 re-implementation

Peartes:
Due to different implementation of the type="time" input ton various browsers and the unwanted consequence to the UI, I am 
redesigning the UI to receive the user's time input using 2 divs and numeric input.
    
The issue arises here because the numeric-input component allows for any length of string (as it should) but time inputs should be limited to only 2 digits each for the Hour and Min interval. Now, the numeric-input component as been given 
another property called maxlength which limits the number of digit that component could receive. To make sure that other
 numeric-input components are not limited to 0 digits, the default length was set at a ridiculous number (better 
implementation is welcomed)

This PR is for the issue [issue](https://www.notion.so/curvelabs/339a249185e14ddcbd731e5ba07c46c7?v=f7cba822bb9043a5a87ab80b5e2b40ef&p=09a5d393720747aeb0c3a8b5bf51ca85) @dkent600


dkent600 (changes requested):
This is way more complicated than it needs to be.  We can simply use a single input with a helpful placeholder asking them to enter the time in GMT as HH:MM in 24-hour time (like 21:00).

Peartes:
I am closing this PR due to a change in implementation following the sync had with @renc-curvelabs and @dkent600

PR #33, opened 2021-05-26
Make time input simple input box with validation

Peartes:
As discussed in our sync @dkent600 @renc-curvelabs @BartuCulha the time input found.  on the 4th stage of seed, registration has been a simple input box with a placeholder describing the format of the time and validation in place to catch wring usage. The issue referred to is [here](https://www.notion.so/curvelabs/339a249185e14ddcbd731e5ba07c46c7?v=f7cba822bb9043a5a87ab80b5e2b40ef&p=09a5d393720747aeb0c3a8b5bf51ca85)


dkent600 (changes requested):
Validation should confirm that the hours are between 0 and 23 and minutes are between 0 and 59

src/newSeed/stage4.ts
```diff
@@ -71,15 +71,21 @@ export class Stage4 extends BaseStage {
       message = "Please select a Start Date";
     } else if (!this.startTime) {
       message = "Please enter a value for the Start Time";
-    } else if (!(Number.parseInt(this.startTime.split(":")[0]) < 25)
-      || !(Number.parseInt(this.startTime.split(":")[1]) < 61)) {
+    } else if (!(Number.parseInt(this.startTime.split(":")[0]) >= 0)
+      || !(Number.parseInt(this.startTime.split(":")[0]) < 24)) {
+      message = "Please enter a valid value for Start Time";
+    } else if (!(Number.parseInt(this.startTime.split(":")[1]) >= 0)
+      || !(Number.parseInt(this.startTime.split(":")[1]) < 60)) {
```
dkent600, line 78: what if they enter a negative number for hours or minutes?

src/newSeed/stage4.ts
```diff
@@ -71,15 +71,21 @@ export class Stage4 extends BaseStage {
       message = "Please select a Start Date";
     } else if (!this.startTime) {
       message = "Please enter a value for the Start Time";
-    } else if (!(Number.parseInt(this.startTime.split(":")[0]) < 25)
-      || !(Number.parseInt(this.startTime.split(":")[1]) < 61)) {
+    } else if (!(Number.parseInt(this.startTime.split(":")[0]) >= 0)
```
dkent600, line 74: you should only execute each split once.

dkent600, line 74: what happens if there is no ':'?

or what happens if the value is not a number (parseInt returns NaN)

dkent600 (changes requested):

src/newSeed/stage4.ts
Peartes: Validation would flag it

src/newSeed/stage4.ts
Peartes: Validation flags it

Peartes: Validation catches all corner cases I tested but please in case you think of any other

Peartes: Done

dkent600 (approved):

PR #34, opened 2021-05-26
Make checkbox div hot

Peartes:
This commit addresses the issue [here](https://www.notion.so/curvelabs/339a249185e14ddcbd731e5ba07c46c7?v=f7cba822bb9043a5a87ab80b5e2b40ef&p=8d68ec3dcd184a67837e84277aded8c0) where only the label on the checkbox is clickable


src/newSeed/stage4.ts
```diff
@@ -34,6 +34,14 @@ export class Stage4 extends BaseStage {
     });
   }
 
+  toggleCheckbox(name: string): void {
+    if (name === "whitelist") {
+      this.seedConfig.seedDetails.whitelist.isWhitelist = !this.seedConfig.seedDetails.whitelist.isWhitelist;
+    } else {
+      this.seedConfig.seedDetails[name] = !this.seedConfig.seedDetails.geoBlock;
```
dkent600, line 41: ass this line is written, `[name]` will not work for multiple cases.  It is misleadingly generalized.  I would just explicitly use `.geoBlock` instead of `[name]`

src/newSeed/stage4.html
```diff
@@ -130,7 +130,7 @@
               <question-mark text="Tick"></question-mark>
             </div>
           </div>
-          <div class="checkboxContainer">
+          <div class="checkboxContainer" click.trigger="toggleCheckbox('whitelist')">
```
dkent600, line 133: you can't use `delegate` instead of `trigger`?

dkent600 (changes requested):

dkent600 (approved):

PR #36, opened 2021-05-27
Cleanup the homepage as requested from QA

Peartes:
This commit addresses this issues: [issue1](https://www.notion.so/curvelabs/339a249185e14ddcbd731e5ba07c46c7?v=f7cba822bb9043a5a87ab80b5e2b40ef&p=c827e137941344a28a47af47e4c2da9d) [issue2](https://www.notion.so/curvelabs/339a249185e14ddcbd731e5ba07c46c7?v=f7cba822bb9043a5a87ab80b5e2b40ef&p=9489723aceda485cbaf191c8660528c2) [issue3](https://www.notion.so/curvelabs/339a249185e14ddcbd731e5ba07c46c7?v=f7cba822bb9043a5a87ab80b5e2b40ef&p=892cdf0bd40444999ec18d4a62faa014)


src/home/home.scss
```diff
@@ -55,7 +55,7 @@
       > .subtitle {
         font-family: "Inter";
         font-weight: bold;
-        font-size: 1.3rem;
+        font-size: 1.4rem;
```
dkent600, line 58: can we please use the pixel size given by the mock?

src/home/home.html
```diff
@@ -28,7 +28,7 @@
     </div>
     <div class="section launches">
       <template if.to-view="featuredSeeds.length">
-        <div class="title heading heading1">Featured Launches</div>
+        <div class="title heading heading3">Featured Launches</div>
```
dkent600, line 31: please just override the css for heading1 rather than creating in home.scss a heading3 (which also exists in styles.scss)

src/home/home.html
```diff
@@ -51,7 +51,7 @@
     <div class="section fairLaunch">
       <div class="pic"><img src="/spacex-OHOU-5UVIYQ-unsplash.png" /></div>
       <div class="rightColumn">
-        <div class="title heading heading1">The real fair launch</div>
+        <div class="title heading heading3">The real fair launch</div>
```
dkent600, line 54: please just override the css for heading1 rather than creating in home.scss a heading3 (which also exists in styles.scss)

src/home/home.scss
```diff
@@ -119,6 +119,10 @@
         text-align: center;
         padding: 2em;
       }
+      & .heading3 {
```
dkent600, line 122: you can override heading1 (from styles.scss) here.  Note it is possible that the lineheight is actually what you want to change here, not the margin.  Check in the mock.

src/home/home.scss
```diff
@@ -174,6 +178,11 @@
       column-gap: 10%;
       font-family: Inter;
 
+      & .heading3 {
```
dkent600, line 181: same here re heading1

dkent600 (changes requested):

src/home/home.html
Peartes: I thought the &.heading3 makes sure to only style any heading3 that is a child of the parent element

Peartes: I thought heading3 might be better because headin1 has a gradient as it background that might require more CSS to change it style and heading3 only needs to change the font-size

src/home/home.scss
```diff
@@ -16,7 +16,7 @@
     &.fairLaunch {
       margin-left: 4.5em;
       margin-right: 4.5em;
-      margin-bottom: 5.75em;
+      margin-bottom: 8em;
```
dkent600, line 19: is there a px value we can use from the wireframe?

dkent600 (changes requested):

src/home/home.scss
dkent600: same for the margins

Peartes: Maybe a PR to change some values to their px values because some component also uses this particular values. By the way do you want to remove all uses of em and rem or just for margins and paddings ?

dkent600: For now just do it when we're in the code and looking at the wireframe, like this code here.

dkent600: I think we should us px for everything, frankly

PR #37, opened 2021-05-28
Clean-up footer to match spec

Peartes:
This addresses a bunch of issues on the footer. The issues are here [issue](https://www.notion.so/curvelabs/339a249185e14ddcbd731e5ba07c46c7?v=f7cba822bb9043a5a87ab80b5e2b40ef&p=9de820ea6ca840bab51b2dce870c176f)


dkent600 (changes requested):
This is changes a bunch of stuff that isn't in the given card.  The card only asks to change the indentation.  Since this is a global template issue I should work it (since I've kinda taken ownership of that).  The rest of the stuff should be worked in correlation with individual cards, one card at a time.

PR #38, opened 2021-05-31
Align the no-miss section to match the homepage

Peartes:
This commit addresses the issue [here](https://www.notion.so/curvelabs/339a249185e14ddcbd731e5ba07c46c7?v=f7cba822bb9043a5a87ab80b5e2b40ef&p=d9cf6ed33385477195f57f6d3514ad26)


src/home/home.scss
```diff
@@ -127,7 +127,7 @@
       background-color: $BG02;
       height: 12em;
       align-content: center;
-      justify-content: space-around;
+      justify-content: space-between;
```
dkent600, line 130: This does not result in what the wireframe is asking for.

dkent600 (changes requested):

src/home/home.scss
Peartes: Screenshot 2021-05-31 at 16 35 47

Here is the result of the change

dkent600: take a thorough look  at the home page wireframe.  if you still don't see the problem, set up a zoom for us and you can show me.

Peartes:
@dkent600  work on homepage would take care of this

PR #40, opened 2021-06-01
Add documentation pages with dummy data

Peartes:
This commit adds the documentation pages with dummy data addressing the issue [here](https://www.notion.so/curvelabs/339a249185e14ddcbd731e5ba07c46c7?v=f7cba822bb9043a5a87ab80b5e2b40ef&p=f8d4d82cd2654325a4e47194b84501bd)


src/documentation/documentation.html
```diff
@@ -2,10 +2,12 @@
   <div class="page animated-page au-animate documentsContainer">
     <div class="sidebar">
       <template repeat.for="doc of router.navigation">
-        <a href.to-view="doc.href"
-          class="document ${doc.isActive ? 'active' : ''}">
-          ${doc.title}
-        </a>
+        <div>
```
dkent600, line 5: better than adding an unnecessary div is to add `display: block` to the `a` (in the scss)

dkent600 (changes requested):

src/documentation/documentation.scss
```diff
@@ -1,2 +1,38 @@
+@import "styles/styles.scss";
 .documentsContainer {
```
dkent600, line 2: The layout of the page does not conform to the doc.  (For example: Is missing the header)

dkent600: still don't have it

image

Peartes: This is what the branch currently has @dkent600 

screenshot

Peartes: This is what the branch currently has 
Screenshot 2021-06-02 at 10 37 07

Peartes: The layout now matches. Please re-review

dkent600: Doesn't match.  There should not be a different background color and there is too much column whitespace between "Documentation" and "Title 1"

image

Peartes: Okay so here is the adjusted documentation page, should match the mocks now.
Screenshot 2021-06-02 at 12 51 06

src/documentation/documentation.scss
```diff
@@ -1,2 +1,43 @@
...
+            text-decoration: none;
+            display: block;
+        }
+
+        .active {
+            color: $Secondary02;
+        }
+    }
+    .docsContainer {
+        background-color: $BG01;
```
dkent600, line 27: why is this needed?

dkent600 (approved):

PR #45, opened 2021-06-02
Stage7 close button

Peartes:
Add a cancel button tot he stages 7 of the seed reg wizard and redirects back to the home page when the button is clicked.

The issue addressed is [There should be a "Close" button in the "Thank You" 7th Stage](https://www.notion.so/curvelabs/339a249185e14ddcbd731e5ba07c46c7?v=f7cba822bb9043a5a87ab80b5e2b40ef&p=80c9dc8930bd408b99c8d0cae223e73a)


dkent600 (changes requested):
Can you please reduce the amount of space between the text and the button? Way too much whitespace. Make it more nicely balanced in that respect.  Thanks.

image

dkent600 (approved):

PR #54, opened 2021-06-07
Launch page cleanup

Peartes:
This simple commit addresses this issue found [here](https://www.notion.so/curvelabs/339a249185e14ddcbd731e5ba07c46c7?v=f7cba822bb9043a5a87ab80b5e2b40ef&p=9aa6f6dbed4e4ef88aff35a764fdfe3c)


src/launch/launch.html
```diff
@@ -6,10 +6,8 @@
         <div>Apply to host your Launch</div>
       </div>
       <div class="subtext">
-        <div>
-          <div>PrimeLAUNCH empowers and creators to host and fair and</div>
-          <div>seemless launch in a completely permissionless setting.</div>
-        </div>
+        PrimeLAUNCH empowers builders and creators to host a fair and
+        <br /> seemless launch in a completely permissionless setting.
```
dkent600, line 10: Please remove this BR, nor split with divs.  This is not a good practice except under special circumstances.  Just let the text flow as it will.  Shrink the container if you must.

dkent600 (changes requested):

src/launch/launch.html
Peartes: Okay

src/launch/launch.scss
```diff
@@ -22,9 +22,15 @@
...
     font-weight: normal;
-    font-size: 0.85em;
+    font-size: 18px;
     color: $Neutral03;
   }
+
+  .banner > .subtext {
+    font-weight: bold;
+    text-align: center;
+    width: 615px;
```
dkent600, line 32: did you test this at various widths?  Would probably be better as % or [n]vw

dkent600 (changes requested):

src/launch/launch.scss
Peartes: Yes it works at various width of the viewport but I can also make it em, % or vw also

Peartes: Okay so I tried this with various viewport width and it is fine, should I leave as is or change to some relative width value?

dkent600 (approved):

PR #56, opened 2021-06-08
Wizard config state

Peartes:
There is another bug with having a navigable seed reg wizard using the sidebar which is that the proceed function of some stages mainly 3 and 4 usually persist some data into the see config state after validation is performed but if the user navigates away using the sidebar, the proceed function would never be called therefore the data to be persisted are never persisted.
    
 This commit takes this approach at fixing the problem: 
Simply calling the proceed function introduces an unwanted feature of validation error for a 
stage navigated away from being displayed in the stage navigated to because the 
validation error component for a stage is triggered in the stage's proceed function instead a 
new function called persisted data() is introduced into the view-model for the stages that require persisting data after validation. This persistData function is invoked
 by the proceed function after validation is complete. This function is also invoked by the detached method of each class extending the abstract base stage class to persist data after validation has been complete.

This commit closes this [issue](https://www.notion.so/curvelabs/339a249185e14ddcbd731e5ba07c46c7?v=f7cba822bb9043a5a87ab80b5e2b40ef&p=3339e916f8654d67ab5d420843b106ee)


src/newSeed/stage3.ts
```diff
@@ -33,13 +33,18 @@ export class Stage3 extends BaseStage {
       this.validationError(message);
       this.stageState.verified = false;
     } else {
-      this.stageState.verified = true;
-      this.wizardState.seedTokenSymbol = this.seedSymbol;
-      this.wizardState.fundingTokenSymbol = this.fundingSymbol;
+      this.persistData();
```
dkent600, line 36: this is not necessary since persistData is called in detached

src/newSeed/stage3.ts
```diff
@@ -33,13 +33,18 @@ export class Stage3 extends BaseStage {
...
       this.next();
     }
   }
 
+  persistData(): boolean {
+    this.stageState.verified = true;
```
dkent600, line 42: this line should stay in `proceed`

src/newSeed/stage4.ts
```diff
@@ -53,26 +53,31 @@ export class Stage4 extends BaseStage {
...
+    // Get the start and end time
+    const startTimes = this.startTime.split(":");
+    const endTimes = this.endTime.split(":");
+    let temp = this.startDate;
+    temp.setHours(Number.parseInt(startTimes[0]), Number.parseInt(startTimes[1]));
+    this.seedConfig.seedDetails.startDate = this.dateService.toISOString(this.dateService.translateLocalToUtc(temp));
+    temp = this.endDate;
+    temp.setHours(Number.parseInt(endTimes[0]), Number.parseInt(endTimes[1]));
+    this.seedConfig.seedDetails.endDate = this.dateService.toISOString(this.dateService.translateLocalToUtc(temp));
+    this.stageState.verified = true;
```
dkent600, line 77: this line should stay in `proceed`

dkent600 (changes requested):

src/newSeed/stage4.ts
```diff
@@ -53,26 +53,31 @@ export class Stage4 extends BaseStage {
...
-      const startTimes = this.startTime.split(":");
-      const endTimes = this.endTime.split(":");
-      let temp = this.startDate;
-      temp.setHours(Number.parseInt(startTimes[0]), Number.parseInt(startTimes[1]));
-      this.seedConfig.seedDetails.startDate = this.dateService.toISOString(this.dateService.translateLocalToUtc(temp));
-      temp = this.endDate;
-      temp.setHours(Number.parseInt(endTimes[0]), Number.parseInt(endTimes[1]));
-      this.seedConfig.seedDetails.endDate = this.dateService.toISOString(this.dateService.translateLocalToUtc(temp));
-      this.stageState.verified = true;
+      this.persistData();
```
dkent600, line 61: this is not necessary since persistData is called in detached

src/newSeed/baseStage.ts
```diff
@@ -69,6 +72,10 @@ export abstract class BaseStage {
     return null;
   }
 
+  protected persistData(): boolean {
```
dkent600, line 75: why return boolean?  Should be void

src/newSeed/stage3.ts
```diff
@@ -33,16 +33,13 @@ export class Stage3 extends BaseStage {
       this.validationError(message);
       this.stageState.verified = false;
     } else {
-      this.persistData();
```
dkent600: don't you need ` this.stageState.verified = true;` here?

src/newSeed/stage4.ts
```diff
@@ -58,12 +58,11 @@ export class Stage4 extends BaseStage {
       this.validationError(message);
       this.stageState.verified = false;
     } else {
-      this.persistData();
```
dkent600: Don't you need ` this.stageState.verified = false;` here?

dkent600 (changes requested):

src/newSeed/stage3.ts
Peartes: Not really, it is called by the detach function and set there

Peartes: Not really, it's funny but the detached function also calls the verifyInputs() and persistData() functions and also sets the stage state to true. Since the proceed function also triggers the detached state, the stage state is set after the call to validate input is done in the detached state.

src/newSeed/stage3.ts
```diff
@@ -34,12 +34,15 @@ export class Stage3 extends BaseStage {
       this.stageState.verified = false;
```
dkent600, line 34: I think we should limit all changes to `this.stageState.verified` so they occur only inside of `validateInputs()`

dkent600 (changes requested):

dkent600 (approved):

PR #58, opened 2021-06-08
Homepage cleanup

Peartes:
This simple commit increases the font of the subtext on the homepage and uses divs to achieve a linebreak in the sentence
 This address the issue [here](https://www.notion.so/curvelabs/339a249185e14ddcbd731e5ba07c46c7?v=f7cba822bb9043a5a87ab80b5e2b40ef&p=892cdf0bd40444999ec18d4a62faa014)


src/home/home.html
```diff
@@ -5,8 +5,10 @@
         <img src="/nasa-pd4lo70LdbI-unsplash.png" />
       </div>
       <div class="title heading heading0">Seamlessly fundraise for your moonshot</div>
-      <div class="subtitle">PrimeLAUNCH is a platform for fair token launches in a
-        completely permissionless fashion.</div>
+      <div class="subtitle">
+        <div>PrimeLAUNCH is a platform for fair token</div>
```
dkent600, line 9: Just like in another PR, this is the wrong thing to do.  we should wrap text naturally.  The important thing is that the text wrap well at arbitrary page widths.  Do that by correctly setting the width of the container.

dkent600 (changes requested):

src/home/home.html
Peartes: Okay, I will leave the text to wrap naturally but would not set a width on the div because it would not produce the desired result of having the text wrap once on a particular word

src/home/home.html
```diff
@@ -5,8 +5,9 @@
         <img src="/nasa-pd4lo70LdbI-unsplash.png" />
       </div>
       <div class="title heading heading0">Seamlessly fundraise for your moonshot</div>
-      <div class="subtitle">PrimeLAUNCH is a platform for fair token launches in a
-        completely permissionless fashion.</div>
+      <div class="subtitle">
+        <div>PrimeLAUNCH is a platform for fair token launches in a completely permissionless fashion.</div>
```
dkent600, line 9: why do you need an extra div here?

dkent600 (changes requested):

src/home/home.html
dkent600: Usually it is fine as long as the width of the container is correct in the context of its surroundings. Splitting on a particular word is rarely important, and often not feasible at all browser page widths.

dkent600 (approved):

PR #64, opened 2021-06-14
Funding max validation

Peartes:
The funding max cannot be greater than the multiplication of the maximum seed token supply and the price per token for funding tokens. This commit addresses this issue as raised [here](https://www.notion.so/curvelabs/339a249185e14ddcbd731e5ba07c46c7?v=f7cba822bb9043a5a87ab80b5e2b40ef&p=33e30ce33ba84d53a2ae20a01085deb5) 

Please note that this commit has changed the type of the price per token to a normal number (string form) instead of the wei value it was previously, this is because the wei form of this value appends some number of zeros to this value which increases the value when multiplied with the maximum number of seed tokens available. 

For example, if the max seed token was 20 i.e 20e18 wei, and the price per token is 1 i.e 1e18, and the maximum funding target is 1000 i.e 1e21,

 the multiplication of the max seed and price using BigNumber multiplication would be larger than 20e18 (which should be the cap) introducing a bug in the application. To address this issue, the price per token has been made an ordinary number in the case of the example, it would be 1 (just 1) and the multiplication using BigNumber would be 20e18 which is the right cap


src/newSeed/stage4.ts
```diff
@@ -93,12 +93,14 @@ export class Stage4 extends BaseStage {
       message = "Please enter a non-zero value for the Funding Target";
     } else if (!this.seedConfig.seedDetails.fundingMax || this.seedConfig.seedDetails.fundingMax === "0") {
       message = "Please enter a non-zero number for the Funding Max";
+    } else if (BigNumber.from(this.seedConfig.seedDetails.fundingTarget).gt(this.seedConfig.seedDetails.fundingMax)) {
+      message = "Please enter a value for Funding Target smaller than Funding Max";
+    } else if (this.seedConfig.tokenDetails.maxSeedSupply && BigNumber.from(this.seedConfig.seedDetails.fundingMax).gt(BigNumber.from(this.seedConfig.tokenDetails.maxSeedSupply).mul(this.seedConfig.seedDetails.pricePerToken))) {
```
dkent600, line 98: Can be:

```
this.seedConfig.tokenDetails.maxSeedSupply && BigNumber.from(this.seedConfig.seedDetails.fundingMax).gt(this.seedConfig.tokenDetails.maxSeedSupply.mul(this.seedConfig.seedDetails.pricePerToken))) {
```

src/newSeed/stage4.html
```diff
@@ -9,12 +9,11 @@
         <div class="infoInput">
           <label for="ppt">Funding Tokens per Seed Token</label>
           <div class="pptContainer">
-            <numeric-input
-              output-as-string="true"
+            <input
```
dkent600, line 12: Why did you switch this to input?  I don't see any reason, and it has to be a bignumber in wei.

dkent600 (changes requested):

src/newSeed/stage3.ts
```diff
@@ -49,6 +49,8 @@ export class Stage3 extends BaseStage {
       message = "Please enter a valid address for the Seed Token Address";
     } else if (!this.seedConfig.tokenDetails.maxSeedSupply || this.seedConfig.tokenDetails.maxSeedSupply === "0") {
       message = "Please enter a non-zero number for Maximum Supply";
+    } else if (this.seedConfig.seedDetails.fundingMax && this.seedConfig.seedDetails.pricePerToken && BigNumber.from(this.seedConfig.seedDetails.fundingMax).gt(BigNumber.from(this.seedConfig.tokenDetails.maxSeedSupply).mul(this.seedConfig.seedDetails.pricePerToken))) {
```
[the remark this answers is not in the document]
Peartes: this.seedConfig.tokenDetails.maxSeedSupply is not a BigNumber but a string that is why I had to convert to a BigNumber first and the check for the this.seedConfig.seedDetails.fundingMax and this.seedConfig.seedDetails.pricePerToken is to avoid a crash when converting to BigNumber. The engine complains when trying to convert "undefined" types to BigNumber

src/newSeed/stage3.ts
```diff
@@ -49,6 +50,9 @@ export class Stage3 extends BaseStage {
       message = "Please enter a valid address for the Seed Token Address";
     } else if (!this.seedConfig.tokenDetails.maxSeedSupply || this.seedConfig.tokenDetails.maxSeedSupply === "0") {
       message = "Please enter a non-zero number for Maximum Supply";
+    } else if (this.seedConfig.seedDetails.fundingMax && this.seedConfig.seedDetails.pricePerToken &&
+      Number(fromWei(this.seedConfig.seedDetails.fundingMax)) > Number(fromWei(this.seedConfig.tokenDetails.maxSeedSupply)) * Number(fromWei(this.seedConfig.seedDetails.pricePerToken))) {
```
dkent600, line 54: Our practice everywhere is to convert the bignumber to a string and pass that to `numberService.fromString`.  Can you please follow that here?  Thanks.

src/newSeed/stage4.ts
```diff
@@ -86,19 +87,20 @@ export class Stage4 extends BaseStage {
...
-
     if (!this.seedConfig.seedDetails.pricePerToken || this.seedConfig.seedDetails.pricePerToken === "0") {
       message = "Please enter a value for Funding Tokens per Seed Token";
     } else if (!this.seedConfig.seedDetails.fundingTarget || this.seedConfig.seedDetails.fundingTarget === "0") {
       message = "Please enter a non-zero value for the Funding Target";
     } else if (!this.seedConfig.seedDetails.fundingMax || this.seedConfig.seedDetails.fundingMax === "0") {
       message = "Please enter a non-zero number for the Funding Max";
+    } else if (BigNumber.from(this.seedConfig.seedDetails.fundingTarget).gt(this.seedConfig.seedDetails.fundingMax)) {
+      message = "Please enter a value for Funding Target smaller than Funding Max";
+    } else if (this.seedConfig.tokenDetails.maxSeedSupply && Number(fromWei(this.seedConfig.seedDetails.fundingMax)) > Number(fromWei(this.seedConfig.tokenDetails.maxSeedSupply)) * Number(fromWei(this.seedConfig.seedDetails.pricePerToken))) {
```
dkent600, line 98: Our practice everywhere is to convert the bignumber to a string and pass that to `numberService.fromString`.  Can you please follow that here?  Thanks.

dkent600 (changes requested):

dkent600 (approved):

PR #68, opened 2021-06-16
Stage3 global distribution bug fixes

Peartes:
In the stage3 of the seed config wizard (the seed global distribution), there exists a bug (more like a lack of feature) - that is deleting a distribution row. A delete button has been added to delete rows unneeded.

Also, the validation has a bug-trying to add together 2 BigNumbers where one of them might not be a BigNumber. Now, we only perform that operation after confirming that both values are now BigNumber. The issue addressed is [here](https://www.notion.so/curvelabs/339a249185e14ddcbd731e5ba07c46c7?v=f7cba822bb9043a5a87ab80b5e2b40ef&p=7e527a0f065e45e899104cc57143cf63)


src/newSeed/stage3.ts
```diff
@@ -69,10 +75,11 @@ export class Stage3 extends BaseStage {
         message = `Please enter a number greater than zero for Category ${tokenDistrb.category} Amount`;
       } else if (!(tokenDistrb.lockup > 0)) {
         message = `Please enter a number greater than zero for Category ${tokenDistrb.category} Lock-up`;
+      } else {
+        totalDistribAmount = totalDistribAmount.add(tokenDistrb.amount);
       }
-      totalDistribAmount = totalDistribAmount.add(tokenDistrb.amount);
     });
-    if (!message && totalDistribAmount.gt(this.seedConfig.tokenDetails.maxSeedSupply)) {
+    if (!message && this.seedConfig.tokenDetails.maxSeedSupply && totalDistribAmount.gt(this.seedConfig.tokenDetails.maxSeedSupply)) {
```
dkent600, line 82: this whole block should be in an `else`.  then you'll already know that maxSeedSupply is ok and you don't need this check.

dkent600 (changes requested):

dkent600 (approved):

PR #70, opened 2021-06-17
Launch page mobile version

Peartes:
Okay, the launch page is already a bit responsive save the preset width in the launch boxes made it difficult and the no-wrap property on the headers and subtext. After removing some un-needed divs and preset width, the responsiveness pops out ;). This commit addresses this [issue](https://www.notion.so/curvelabs/339a249185e14ddcbd731e5ba07c46c7?v=f7cba822bb9043a5a87ab80b5e2b40ef&p=bd54c5d280b843f28af95fa774891579)


dkent600 (changes requested):
The horizontal centering of button and powered-by should happen at the same breakpoint where the page goes to a single column

dkent600 (changes requested):
I didn't ask to break into single column sooner.  One shouldn't break out of the preferred layout, which is desktop (two columns), any sooner than absolutely necessary.  I think you should revert that change.

In any case, the switch to single column and the switch to horiz centered are not happening simultaneously. as they should

dkent600 (approved):

PR #79, opened 2021-06-24
Rewording the headers

Peartes:
Fixing this [issue](https://www.notion.so/curvelabs/339a249185e14ddcbd731e5ba07c46c7?v=f7cba822bb9043a5a87ab80b5e2b40ef&p=3d20acff88154ede904b3d6fa5ea658b) which basically rewords the footer components


src/footer/footer.html
```diff
@@ -30,20 +30,20 @@
           </div>
         </div>
         <div class="section launch">
-          <div class="title heading heading4" click.delegate="showLinks('primeLAUNCH')">PrimeLAUNCH<div class="menubutton fas fa-angle-down"></div>
+          <div class="title heading heading4" click.delegate="showLinks('primeLAUNCH')">PRIMELAUNCH<div class="menubutton fas fa-angle-down"></div>
```
dkent600, line 33: should be "PRIME LAUNCH"

dkent600 (changes requested):
I think we're being super inconsistent with the naming.  @renc-curvelabs would be good if you can review this wrt  PrimeLAUNCH vs "Prime Launch".

dkent600 (approved):

PR #80, opened 2021-06-25
display image by input of a valid url

giladt:
Added an image after the `.infoInput` element.
Added validation for broken links.
Added styling for max width/height of 400px.

Tested on both web/mobile views.

### DIFF
--- a/src/newSeed/stage2.ts
+++ b/src/newSeed/stage2.ts
@@ -1,3 +1,5 @@
+import { bindable } from "aurelia-typed-observable-plugin";
+import axios from "axios";
 import { BaseStage } from "newSeed/baseStage";
 import { Utils } from "services/utils";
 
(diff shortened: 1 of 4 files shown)


src/newSeed/stage2.ts
```diff
@@ -26,9 +35,10 @@ export class Stage2 extends BaseStage {
       message = "Please enter a value for Team Description";
     } else if (!Utils.isValidUrl(encodeURI(this.seedConfig.projectDetails.logo))) {
       message = "Please enter a valid URL for Project Logo";
-    }
-    else if (!this.isValidFile(this.seedConfig.projectDetails.logo)) {
+    } else if (!this.isValidFile(this.seedConfig.projectDetails.logo)) {
       message = "Please supply a valid image file type for Project Logo";
+    } else if (!this.isLinkNotBroken(this.seedConfig.projectDetails.logo)) {
+      message = "No image found under the provided Project Logo URL";
```
dkent600, line 41: make it:  "No image found at the provided Project Logo URL"

src/newSeed/stage2.ts
```diff
@@ -7,6 +7,15 @@ export class Stage2 extends BaseStage {
     return re.test(String(file).toLowerCase());
   }
 
+  private isLinkNotBroken(url: string): boolean {
```
dkent600, line 10: I am getting this for a valid url.

image

dkent600: I'm getting the same error for an invalid URL, but I should not be getting an error at all until the user moves on from the page.

image

dkent600 (changes requested):

giladt:
I haven't changed much since the last commit. Tried different approaches, but the problem is that external URLs are blocked by **CORS** rules, and I can not do anything from the front end to avoid it.
I have tried adding headers to the _devServer_ in _webpack.config.js_ but I'm not sure if I did it correctly or not (didn't commit that as it didn't work) and anyway it should be defined in the production environment too, which may imply security issues...

It works correctly with local images - e.g. `/favicon-32x32.png`
or some cdn images (I guess its whitelisted?) - e.g. `https://gblobscdn.gitbook.com/spaces%2F-MH0VrbEUlD_Vg90Gx8g%2Favatar-1599902686246.png`.

dkent600 (approved):

PR #81, opened 2021-06-26
Making stage7 mobile

Peartes:
The width property set on the stage7 wizards makes it scale poorly on mobile devices. This commit address the scalability issues of the stage7 wizard on mobile. Addressing this [issue](https://www.notion.so/curvelabs/339a249185e14ddcbd731e5ba07c46c7?v=f7cba822bb9043a5a87ab80b5e2b40ef&p=f658617c88be42aba981d5b128c7cc0f)


src/newSeed/baseStage.scss
```diff
@@ -530,5 +530,11 @@
         padding-left: 8px;
       }
     }
+
+    &.stage7 {
+      .linksContainer {
+        width: unset;
```
dkent600, line 536: I think the right way to fix this is to delete line  463.  Our bias should always be less code rather than more, to remove unnecessary code.

dkent600 (changes requested):

dkent600 (approved):

PR #82, opened 2021-06-27
Disabling wizard sidebar

Peartes:
Quoting this conversation from [this issue](https://www.notion.so/curvelabs/339a249185e14ddcbd731e5ba07c46c7?v=f7cba822bb9043a5a87ab80b5e2b40ef&p=3c109b4696a54adfa083d46bf57c4248): 

"After successful registration, the user has the ability to click to the circular buttons on the right, and review any information they have written. They also have the ability to resubmit. In order not to confuse the users and not to have seeds duplicated, can we make the circular buttons unclickable within this page? @Renç Korzay @Doug Kent "

Using the event aggregator, we simply publish a seed-created event upon successful seed creation while in the newSeed view-model, we have subscribed to this event and switch a state variable "seedCreated" which is a boolean to true to signify that a seed has been created, this variable determines if we would navigate to a stage when the stage is selected using the side-bar. When the variable is true, we cannot navigate to any stage using the sidebar and vice-versa.


src/newSeed/newSeed.ts
```diff
@@ -19,7 +22,7 @@ export class NewSeed {
     return this.router.currentInstruction.config;
   }
 
-  constructor() {
+  constructor(private eventAggregator: EventAggregator) {
```
dkent600, line 25: Please use wizardState, not events for managing wizard-scoped, non-seedConfig data.

src/newSeed/newSeed.ts
```diff
@@ -162,7 +166,9 @@ export class NewSeed {
   }
 
   setStage(route: string): void {
-    this.router.navigate(route);
+    if (!this.seedCreated) {
+      this.router.navigate(route);
```
dkent600, line 170: how are we supposed to get to stage7?

dkent600: Note this isn't just about navigation.  What about the browser back and forward buttons?

I think you simply need to:

1) clear the seedConfig and wizardData after the Seed creation _transaction_ is created (BTW, note that the Seed itself will not yet have been created at any point in this process).  Do this upon entry to stage 7.
2) hide the menu on stage7.

src/newSeed/newSeed.ts
dkent600: You can set this state upon arrival in Stage 7

dkent600 (changes requested):

src/newSeed/seedConfig.ts
```diff
@@ -99,4 +100,24 @@ export class SeedConfig implements ISeedConfig {
   public seedDetails = {
     geoBlock: false,
   } as ISeedDetails;
+
+  clearState(): void {
+    this.general = {
```
dkent600, line 105: Duplicating code is not a good design.  Make the constructor call this function.

src/newSeed/seedConfig.ts
```diff
@@ -78,6 +78,7 @@ export interface ISeedConfig {
   tokenDetails: ITokenDetails,
   contactDetails: IContactDetails,
   seedDetails: ISeedDetails,
+  clearState: () => void,
```
dkent600, line 81: does this appear in JSON.stringify?

dkent600 (changes requested):

src/newSeed/seedConfig.ts
Peartes: No, a copy of the seedConfig excluding the function is stringified

Peartes: Stringify ignores the function

src/newSeed/seedConfig.ts
```diff
@@ -78,25 +78,38 @@ export interface ISeedConfig {
...
 }
 
 export class SeedConfig implements ISeedConfig {
-  public version = "1.0.0";
```
dkent600, line 84: did this line get lost?

dkent600 (changes requested):

src/newSeed/newSeed.html
```diff
@@ -1,14 +1,14 @@
 <template>
   <require from="../resources/elements/circledNumber/circledNumber.html"></require>
 
-  <div class="newSeedContainer">
-    <div class="sideBarDropDownContainer">
+  <div class="newSeedContainer ${currentStage.settings.stageNumber === 7 ? 'stage7' : ''}">
```
dkent600, line 4: can you please put the `router.currentInstruction.config.name` as the class name for every seed, rather then special casing the seventh?

src/newSeed/newSeed.html
```diff
@@ -1,14 +1,14 @@
...
+    <div class="sideBarDropDownContainer" show.to-view="currentStage.settings.stageNumber === 7 ? false : true">
```
dkent600, line 5: Can you please just hide this in .css to eliminate this special casing in code?

src/newSeed/newSeed.html
```diff
@@ -1,14 +1,14 @@
...
       <div click.delegate="toggleSideBar()" class="sideBarDropDown">
         ${currentStage.title}
         <div class="fas fa-angle-down"></div>
       </div>
     </div>
-    <div class="currentStageSidebar" ref="sideBar">
+    <div class="currentStageSidebar" ref="sideBar" show.to-view="currentStage.settings.stageNumber === 7 ? false : true">
```
dkent600, line 11: Can you please just hide this in .css to eliminate this special casing in code?

dkent600 (changes requested):

src/newSeed/newSeed.html
```diff
@@ -1,14 +1,14 @@
 <template>
   <require from="../resources/elements/circledNumber/circledNumber.html"></require>
 
-  <div class="newSeedContainer ${currentStage.settings.stageNumber === 7 ? 'stage7' : ''}">
-    <div class="sideBarDropDownContainer" show.to-view="currentStage.settings.stageNumber === 7 ? false : true">
+  <div class="newSeedContainer ${currentStage.name}">
+    <div class="sideBarDropDownContainer  ${currentStage.name}">
```
dkent600, line 5: You should needn't this class.  Is enough to it have the class on the containing element

src/newSeed/newSeed.html
```diff
@@ -1,14 +1,14 @@
...
       <div click.delegate="toggleSideBar()" class="sideBarDropDown">
         ${currentStage.title}
         <div class="fas fa-angle-down"></div>
       </div>
     </div>
-    <div class="currentStageSidebar" ref="sideBar" show.to-view="currentStage.settings.stageNumber === 7 ? false : true">
+    <div class="currentStageSidebar ${currentStage.name}" ref="sideBar">
```
dkent600, line 11: You should needn't this class. Is enough to it have the class on the containing element (newSeedContainer)

dkent600 (changes requested):

src/newSeed/newSeed.scss
```diff
@@ -61,7 +61,12 @@
     grid-template-columns: unset;
     .sideBarDropDownContainer {
       display: block;
-      &.stage7 {
+    }
+    &.stage7 {
+      .currentStageSidebar {
```
dkent600, line 66: seems like you don't need this, was already done above

dkent600 (changes requested):

dkent600 (approved):

PR #96, opened 2021-07-05
Show seed admin address

Peartes:
Display the seed admin address instead of the "you" in the stages 7 and 8 of the wizard. This addresses this [issue](https://www.notion.so/curvelabs/339a249185e14ddcbd731e5ba07c46c7?v=f7cba822bb9043a5a87ab80b5e2b40ef&p=956b685b35f941f09bf91fcc8d500783)


src/newSeed/baseStage.ts
```diff
@@ -21,6 +21,7 @@ export interface IWizardState {
   requiredSeedDeposit?: number;
   requiredSeedFee?: number;
   primeDaoAddress?: Address;
+  seedAdminAddress?: Address;
```
dkent600, line 24: why add this? why not just use `seedConfig.seedDetails.adminAddress`?

dkent600 (changes requested):

src/newSeed/baseStage.ts
Peartes: Stage8 also displays the seed admin address, but the seedConfig state gets cleared in stage7 upon creating a seed. So, I added that information to the wizard state to allow it to persist until stage8.

src/newSeed/stage4.ts
```diff
@@ -82,6 +82,7 @@ export class Stage4 extends BaseStage {
     temp = this.endDate;
     temp.setHours(Number.parseInt(endTimes[0]), Number.parseInt(endTimes[1]));
     this.seedConfig.seedDetails.endDate = this.dateService.toISOString(this.dateService.translateLocalToUtc(temp));
+    this.wizardState.seedAdminAddress = this.seedConfig.seedDetails.adminAddress;
```
dkent600, line 85: can you please add a comment here to explain why this is needed?  Thanks

dkent600 (changes requested):

src/newSeed/stage4.ts
Peartes: Okay will do

dkent600 (approved):

PR #98, opened 2021-07-06
Changing stage4 tooltip texts

Peartes:
Simply rewording some tool tips for clarty. This commit addresses this [issue](https://www.notion.so/curvelabs/339a249185e14ddcbd731e5ba07c46c7?v=f7cba822bb9043a5a87ab80b5e2b40ef&p=09a4703aadd341cb84d0dd77fff01998)


src/newSeed/stage4.html
```diff
@@ -149,7 +149,7 @@
           <div class="labeledQuestion">
             <label for="disclaimer">Legal Disclaimer</label>
             <div class="questionMark">
-              <question-mark text="If supplied, all contributors must accept the disclaimer in order to access the launch. The disclaimer must be formatted as plain text or markdown"></question-mark>
+              <question-mark text="If supplied, all contributors must accept the disclaimer in order to access the launch. The disclaimer must be formatted as plain text or markdown and should be in .xx or .xx formats to load correctly"></question-mark>
```
dkent600, line 152: Pat didn't mean for us to literally say "should be in .xx or .xx formats"

dkent600: In this case, just delete that part of the sentence

dkent600 (changes requested):
I'm rejecting this entire PR.  The text is already good enough, and adding "should be in .xx or .xx formats" is just plain wrong, not what Pat intended.  We should not work this one.   Just close it.

Peartes:
Closing this PR because the text reworded was already good enough.

PR #99, opened 2021-07-06
Adding footer logo

Peartes:
We had originally removed the Prime Dao logo from the footer which is an undesied change, this commit adds the logo back and closes this [issue](https://www.notion.so/curvelabs/339a249185e14ddcbd731e5ba07c46c7?v=f7cba822bb9043a5a87ab80b5e2b40ef&p=f16f086f74274cedb89f752a41eefd8c)


src/footer/footer.html
```diff
@@ -53,6 +53,7 @@
 
     <div class="bottom">
       <div class="copyright"><img src="/cc.svg">2021</div>
```
dkent600, line 55: I don't see 2021 in the screen shot.  I also don't see the copyright icon.  Aren't you supposed to get rid of those?

dkent600 (changes requested):

src/footer/footer.html
Peartes: Sorry, I don't understand. What screenshot? The 2021 icon and copyright icon are still there, I did not get rid of them.

dkent600: The screenshot in the notion card that shows what this PR is supposed to be doing

dkent600 (changes requested):
Isn't looking right on small widths.  (c) is gone, "Governed By" is still visible, and the text is scrunched.

image

dkent600 (changes requested):
logo is missing at wide widths

image

dkent600 (changes requested):
I think this will look a lot better, and Pat will like it a lot better, if we make the (c) Prime Dev... text be centered above a line that contains only Privacy Policy and Cookies.  This would only be needed at the widths where the logo moves to the top and  Privacy Policy and Cookies move to the bottom.  Can you please make that change?

dkent600 (approved):

PR #104, opened 2021-07-08
Seed summary modifications

giladt:
- [Fixed the padding for `.summaryContainer`](https://www.notion.so/curvelabs/Seed-Summary-Increase-the-padding-between-sections-7cfbda2d338f40e1bde898a81ee61ba2)

- [Modified text changes](https://www.notion.so/curvelabs/Seed-Summary-Text-Changes-15b657f77f0b49408dc963d2a2501a2a)

### DIFF
--- a/src/newSeed/baseStage.scss
+++ b/src/newSeed/baseStage.scss
@@ -540,7 +540,7 @@
           }
         }
 
-        + * + * {
+        &:not(:first-child) {
           margin-top: 40px;
         }
       }
@@ -596,7 +596,7 @@
       margin-bottom: 30px;
       word-break: break-word;
 
-      + * + * {
+      &:not(:first-child) {
         margin-top: 30px;
       }
     }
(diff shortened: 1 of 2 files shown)


src/newSeed/baseStage.scss
```diff
@@ -539,7 +539,12 @@
             background-color: $Shade03;
           }
         }
+
+        + * + * {
```
dkent600, line 543: what does this selector do?

dkent600 (changes requested):

src/newSeed/baseStage.scss
giladt: Same as `:not(:first-child)` but shorter...

dkent600: shorter is not always better.  can we please use the long version since it is very clear in meaning?

giladt: done.

dkent600 (approved):

PR #105, opened 2021-07-08
added background image

giladt:
I've added the background image, exported from Figma. As there's no explicit instruction, I found it make sense to hide it form the mobile view.


dkent600 (approved):
@giladt Going forward, can you please as a general practice include in PR descriptions a link to the card you are addressing and word the title of the PR more specifically than this one?  Thanks.

PR #106, opened 2021-07-09
Stage4 seed date validation

Peartes:
In the case where the start and end dates are similar, we must check the start and end times of the project to make sure the start and are similar or malformed. This commit makes surer of that and closes this [issue](https://www.notion.so/curvelabs/339a249185e14ddcbd731e5ba07c46c7?v=f7cba822bb9043a5a87ab80b5e2b40ef&p=5e132584dc464e0caa2df1afe68f2bd4)


src/newSeed/stage4.ts
```diff
@@ -76,10 +76,11 @@ export class Stage4 extends BaseStage {
     // Get the start and end time
     const startTimes = this.startTime.split(":");
     const endTimes = this.endTime.split(":");
-    let temp = this.startDate;
+    const temp:Date = undefined;
+    Object.assign(temp, this.startDate);
```
Peartes, line 80: I realized that assigning the temp to the start date changes the start time and end date object since the reference is only passed during the assignment. This change to the object affects the object and makes it difficult to appropriately check start and end times.

src/newSeed/stage4.ts
```diff
@@ -76,10 +76,11 @@ export class Stage4 extends BaseStage {
     // Get the start and end time
     const startTimes = this.startTime.split(":");
     const endTimes = this.endTime.split(":");
-    let temp = this.startDate;
+    const temp:Date = undefined;
```
dkent600, line 79: please just make this `let temp = Object.assign({}, this.startDate)`

src/newSeed/stage4.ts
```diff
@@ -76,10 +76,11 @@ export class Stage4 extends BaseStage {
...
+    Object.assign(temp, this.startDate);
     temp.setHours(Number.parseInt(startTimes[0]), Number.parseInt(startTimes[1]));
     this.seedConfig.seedDetails.startDate = this.dateService.toISOString(this.dateService.translateLocalToUtc(temp));
-    temp = this.endDate;
+    Object.assign(temp, this.endDate);
```
dkent600, line 83: please make this `temp = Object.assign({}, this.endDate)`

src/newSeed/stage4.ts
```diff
@@ -140,6 +141,13 @@ export class Stage4 extends BaseStage {
       message = "Please enter a valid value for End Time";
     } else if (this.endDate < this.startDate) {
       message = "Please select an End Date greater than the Start Date";
+    } else if (this.endDate.getTime() === this.startDate.getTime()) {
```
dkent600, line 144: Can you not replace all this with `if (this.endDate <= this.startDate)` ?

dkent600 (changes requested):

src/newSeed/stage4.ts
Peartes: No, it will not work. The endDate and startDate don't have times in them.

dkent600: Sorry, I meant:  `if (this.seedConfig.seedDetails.endDate <= this.seedConfig.seedDetails.startDate)`.

Why not this?

dkent600 (changes requested):

src/newSeed/stage4.ts
Peartes: seedConfig's end and start Date are only set after the validation has been done and the persis data function is called. So their value is undefined at first during validation

dkent600: The nice clean way to do this would be to put this logic into a function and use it in both places.  The function would return the Date object.

```
let temp = this.startDate;
let temp = Object.assign({}, this.startDate);
temp.setHours(Number.parseInt(startTimes[0]), Number.parseInt(startTimes[1]));
```

dkent600 (changes requested):

src/newSeed/stage4.ts
```diff
@@ -71,17 +71,29 @@ export class Stage4 extends BaseStage {
...
     this.seedConfig.seedDetails.startDate = this.dateService.toISOString(this.dateService.translateLocalToUtc(temp));
-    temp = this.endDate;
+    return new Date(this.seedConfig.seedDetails.startDate);
+  }
+
+  setSeedConfigEndDate(): Date {
+    // Set the ISO time
+    // Get the start and end time
+    const endTimes = this.endTime.split(":");
+    const temp = this.startDate;
```
dkent600, line 88: really?  have you tested this?

dkent600 (changes requested):

src/newSeed/stage4.ts
Peartes: Sorry made a mistake cleaning up, all corrected now

dkent600 (approved):

PR #107, opened 2021-07-09
Fix gap between image and text

giladt:
- Gave the text full width, to eliminate the white-space between text and image as suggested in this [issue](https://www.notion.so/curvelabs/Close-the-Gap-between-Fair-Launches-Image-Text-1ce5060974984fdfa98849c49df3eb4c)

- Though not explicitly requested in Figma, I think it makes sense to add a border-radius to the image to keep it consistent with the rest of the elements/design.

### DIFF
--- a/src/home/home.scss
+++ b/src/home/home.scss
@@ -175,15 +175,16 @@
       column-gap: 10%;
       font-family: Inter;
 
-      // .pic {
-      //   img {
-      //     width: 450px;
-      //     height: 450px;
-      //   }
-      // }
+      .pic {
+        img {
+            border-radius: 6px;
+          // width: 450px;
+          // height: 450px;
+        }
+      }
 
       .rightColumn {
-        width: 32em;
+        width: 100%;
         > .heading1 {
           -webkit-text-fill-color: $Secondary02;
         }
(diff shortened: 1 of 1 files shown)


src/home/home.scss
```diff
@@ -175,15 +175,16 @@
...
 
-      // .pic {
-      //   img {
-      //     width: 450px;
-      //     height: 450px;
-      //   }
-      // }
+      .pic {
+        img {
+            border-radius: 6px;
```
dkent600, line 180: I can't approve this change.  It is too significant a difference from what came from the designers.  Feel free to create a card on this for Renc to consider (show him a screenshot of the curvy version).  But please remove this change from this PR.

src/home/home.scss
```diff
@@ -175,15 +175,16 @@
...
+          // width: 450px;
+          // height: 450px;
+        }
+      }
 
       .rightColumn {
-        width: 32em;
+        width: 100%;
```
dkent600, line 187: can you please try instead of this changing line 74 to:

`grid-template-columns: auto 1fr;`

this is the first choice for where to control the layout.  Be sure to check how the mobile sizes behave with this change.

dkent600 (changes requested):

dkent600 (approved):

PR #108, opened 2021-07-09
Seed registration modifications

giladt:
- Aligned the question mark vertically to the center- [issue](https://www.notion.so/curvelabs/Seed-Reg-Stage-2-Question-Marks-should-be-centered-7b0a85fa682049a88e9f52135c9cfc77)
- Added `Optional` place holder to the text area in stage 5 - [issue](https://www.notion.so/curvelabs/Seed-Reg-make-placeholders-consistent-in-wording-wrt-optional-31472724fbfa43d68e5ce4a6c95a2bcb)
- Added 12px padding inside text areas also from the top and bottom to keep it visually persistent with the other input boxes: 

Screenshot 2021-07-10 at 10 57 26 Screenshot 2021-07-10 at 10 59 26

### DIFF
--- a/src/newSeed/baseStage.scss
+++ b/src/newSeed/baseStage.scss
@@ -579,7 +579,6 @@
       }
       .labelContainer {
         width: unset;
-        margin-bottom: 5px;
       }
     }
     .labeledQuestion {
--- a/src/newSeed/baseStage.scss
+++ b/src/newSeed/baseStage.scss
@@ -582,6 +582,9 @@
         margin-bottom: 5px;
       }
     }
+    .labeledQuestion {
+      margin-bottom: 5px;
+    }
     > .heading.heading2 {
       font-size: 26px;
     }
--- a/src/newSeed/baseStage.scss
+++ b/src/newSeed/baseStage.scss
@@ -579,6 +579,7 @@
       }
       .labelContainer {
         width: unset;
+        margin-bottom: 5px;
       }
     }
     > .heading.heading2 {
@@ -586,7 +587,6 @@
     }
     label {
       line-height: 20px;
-      margin-bottom: 5px;
     }
     .summaryContainer {
       margin-bottom: 30px;
--- a/src/newSeed/baseStage.scss
+++ b/src/newSeed/baseStage.scss
@@ -476,7 +476,6 @@
     .infoContainer {
       & textarea {
         height: 7em;
-        padding: 12px;
       }
     }
   }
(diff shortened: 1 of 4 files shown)


src/resources/elements/questionMark/questionMark.scss
```diff
@@ -13,6 +13,5 @@
     align-items: center;
     height: 100%;
     position: relative;
```
dkent600, line 15: I think you can remove this line too

src/newSeed/baseStage.scss
```diff
@@ -476,6 +476,7 @@
     .infoContainer {
       & textarea {
         height: 7em;
+        padding: 12px;
```
dkent600, line 479: this looks good, but why only for this one page?  can you please make this apply to all the textareas in the app, not just this one page?  That means applying the padding in styles.scss instead of here.

dkent600 (changes requested):
Can you please  be more specific here in the PR description so folks like me and QA can know what you did?  thanks

> Added padding also from top and bottom to keep persistent with the other input boxes.

src/newSeed/baseStage.scss
giladt: Fixed that in the styles.scss (and removed the declaration from baseStage.scss).

src/resources/elements/questionMark/questionMark.scss
giladt: Removed.
There is some other small issue I've noticed- On tiny displays, the labels have a `margin-bottom` of 5px, whereas the questionMark does not. It causes the questionMark not to be inlined with the label.

Screenshot 2021-07-10 at 10 49 46

should I open a card for that?

dkent600: there is already a card on that, thanks

dkent600: Hmm, wait, this isn't this essentially the problem this this PR is supposed to be fixing?  Or perhaps your code change has actually exacerbated this one?

lets go ahead and fix it here.

dkent600 (changes requested):
Fix remaining issues

dkent600 (changes requested):
still doesn't look good

image

giladt:
Fixed stage 3. I'm not sure though, why not reuse either `labelContainer` or `labeledQuestion` selectors for all labels with question marks.

dkent600 (approved):

PR #109, opened 2021-07-12
Fix gap between image and text

giladt:
@dkent600 - As discussed, I have changed the grid-gap width to a **fixed** value of 5em, instead of 10%.
Part of this [issue](https://www.notion.so/curvelabs/Close-the-Gap-between-Fair-Launches-Image-Text-1ce5060974984fdfa98849c49df3eb4c)


dkent600 (changes requested):
please fix merge conflicts

src/home/home.scss
```diff
@@ -172,7 +172,7 @@
     &.fairLaunch {
       display: grid;
       grid-template-columns: auto 1fr;
-      column-gap: 10%;
+      column-gap: 5em;
```
dkent600, line 175: We are not using em nor rem.  And that you find are not supposed to be there.  We definitely don't want to add more.  Can you please convert to px?

dkent600 (changes requested):

dkent600 (approved):

PR #110, opened 2021-07-12
Displaying image url

Peartes:
Show the project logo URL beside the logo itself. Closes this [issue](https://www.notion.so/curvelabs/339a249185e14ddcbd731e5ba07c46c7?v=f7cba822bb9043a5a87ab80b5e2b40ef&p=c6272110b43b43b88f21cbc156b598b0)


src/newSeed/baseStage.scss
```diff
@@ -528,6 +528,9 @@
               @include tokenIcon;
               height: 19px;
             }
+            &.projectLogo {
+              grid-template-columns: 0.1fr 1fr;
```
dkent600: The horizontal spacing in this element doesn't always look good at various page widths.  The icon isn't vertically centered.

(BTW I note there is another div here without its own class.)

I just committed changes that make the horizontal and vertical spacing/alignment more consistent and appealing.

dkent600 (changes requested):
as part of this PR, please improve the validation msg for the case when there is no project URL entered

image

dkent600 (approved):

PR #114, opened 2021-07-13
fix token name width

giladt:
The grid column has now auto width instead of fixed `fr`.

Removed the `projectLogo` class and moved its declarations to `tokenInfo` since it applies to both _Project Details_ and _Seed Details_.
Although `projectLogo` has now one extra column, I believe it is not an issue.

This fix refers to the following [issue](https://www.notion.so/curvelabs/Section-for-the-Token-Name-should-have-more-width-ec390c6a9dc1444cba141e58845b2622)

### DIFF
--- a/src/newSeed/baseStage.scss
+++ b/src/newSeed/baseStage.scss
@@ -522,8 +522,8 @@
             column-gap: 10px;
             align-items: center;
             width: fit-content;
-            & div {
-              align-self: center;
+            & .etherscanlink {
+              margin-right: 10px;
             }
             & > img {
               @include tokenIcon;
@@ -536,7 +536,6 @@
             background-color: $Shade03;
           }
         }
-
         &:not(:first-child) {
           margin-top: 40px;
         }
--- a/src/newSeed/baseStage.scss
+++ b/src/newSeed/baseStage.scss
@@ -518,20 +518,16 @@
 
           .tokenInfo {
             display: grid;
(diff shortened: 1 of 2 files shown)


src/newSeed/baseStage.scss
```diff
@@ -518,21 +518,17 @@
...
             width: fit-content;
             & div {
               align-self: center;
-              margin-right: 0;
             }
             & > img {
               @include tokenIcon;
               height: 19px;
             }
-            &.projectLogo {
```
dkent600, line 531: 1) the class `tokenInfo` is used in 8 places on this page.  Can you please make sure that the styles are applied properly for all cases?  Rename the styles if you need to.

2) please make sure that the top-level divs under all of the `tokenInfo` have their own proper class, and in the scss under `tokenInfo`, remove the use of `& div`, referring to a proper class instead.

3) can you please add a bit more space here between clipboard button and symbol?  Like this.  Note you can select `.etherscanlink` if needed.

image

dkent600 (changes requested):

src/newSeed/baseStage.scss
giladt: 1. I've controlled it again for all 8 instances of `tokenInfo`.
2. The declaration for the nested `div` is now removed, as it didn't have any effect, hence no need for a new class name.
3. I've added 10px right-margin for `.etherscanlink` if it nested inside `.tokenInfo`.

dkent600 (approved):

PR #117, opened 2021-07-14
External social media link

Peartes:
Some users might not enter an http(s):// while filling out the social media link in stage1 of the seed wizard. This commit adds an // in front of the links that do not contain the "HTTP(s)://". Closes this [issue](https://www.notion.so/curvelabs/339a249185e14ddcbd731e5ba07c46c7?v=f7cba822bb9043a5a87ab80b5e2b40ef&p=2a4ae5dcfd054944a2c45e5be61b21d7)


src/seedDashboard/seedAbout.ts
```diff
@@ -21,6 +21,13 @@ export class SeedAbout{
     ["pdf", "fas fa-file-pdf"],
   ]);
 
+  formatLink(link: string): string {
+    const pattern = /^(?:http(s)?:\/\/)/;
```
dkent600, line 25: what will this do if they have entered "//medium.com"?  Will you end up with "////medium.com"?  

what is they entered "HTTPS://medium.com"?   Will you end up with "HTTPS:////medium.com"?  

More generally:  is there any way this can screw up?

dkent600 (changes requested):

src/seedDashboard/seedAbout.ts
Peartes: The user could not have entered //medium.com as the stage1 validation would flag it as invalid. Likewise, the validation on stage1 also flags HTTPS://medium.com as the regex used for the validation is not cases sensitive

dkent600: > the validation on stage1 also flags HTTPS://medium.com as the regex used for the validation is not cases sensitive

Don't understand.  The question is about what `formatLink` will do when it encounters  HTTPS://medium.com .

dkent600 (changes requested):

src/seedDashboard/seedAbout.ts
Peartes: it would add a // to the link because the regex isn't case sensitive so it would not pick HTTPS://

dkent600 (approved):

PR #120, opened 2021-07-15
removed padding from subtext to align with header

giladt:
- Padding to the left was removed.
- Added top padding to align sub description to the center of the button. 

Closes this [issue](https://www.notion.so/curvelabs/Check-Eligibility-should-be-aligned-with-Subtext-5272a87913ea4c19a809713ec2ee652d)

### DIFF
--- a/src/seedDashboard/seedDashboard.scss
+++ b/src/seedDashboard/seedDashboard.scss
@@ -280,7 +280,6 @@
         .subdescription {
           padding-left: 0px;
           padding-right: 29px;
-          padding-top: 16px;
           padding-bottom: 16px;
           grid-area: subtext;
         }
--- a/src/seedDashboard/seedDashboard.scss
+++ b/src/seedDashboard/seedDashboard.scss
@@ -278,8 +278,10 @@
           justify-self: flex-start;
         }
         .subdescription {
-          padding: 16px 0px;
+          padding-left: 0px;
           padding-right: 29px;
+          padding-top: 16px;
+          padding-bottom: 16px;
           grid-area: subtext;
         }
         .help {
--- a/src/seedDashboard/seedDashboard.scss
+++ b/src/seedDashboard/seedDashboard.scss
@@ -278,9 +278,8 @@
           justify-self: flex-start;
         }
         .subdescription {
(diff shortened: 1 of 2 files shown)


src/seedDashboard/seedDashboard.scss
```diff
@@ -278,9 +278,8 @@
           justify-self: flex-start;
         }
         .subdescription {
-          padding-left: 29px;
+          padding: 16px 0px;
```
dkent600, line 281: Best to make only the minimal change necessary.  Can you please just change padding-left?

dkent600 (changes requested):

src/seedDashboard/seedDashboard.scss
giladt: I needed not only to change the left padding but also add top padding to the `.subDescription` class, it makes the code more readable in my opinion when is shorter and in general more performant.

src/seedDashboard/seedDashboard.scss
```diff
@@ -278,8 +278,9 @@
           justify-self: flex-start;
         }
         .subdescription {
-          padding-left: 29px;
+          padding-left: 0px;
           padding-right: 29px;
+          padding-top: 16px;
```
dkent600, line 283: Why did you add this?  It wasn't called for in the card

dkent600 (changes requested):

src/seedDashboard/seedDashboard.scss
dkent600: The card didn't ask for a change in the top padding did it?

src/seedDashboard/seedDashboard.scss
giladt: I have discussed this with @BartuCulha earlier this morning and got his consent to do that.

giladt: I've added an image and a comment for that in the card, but as I checked it as resolved, now the comment is hidden 😕

dkent600 (approved):
I removed the vertical padding change because of consensus between Renc and me.  And removed a period that didn't belong.

PR #121, opened 2021-07-15
Validate input optimisation

Peartes:
`validateInputs` in the Wizard stages is being invoked twice when the user clicks on the "Next" button.  As we add more stuff to that function (like checking the web for a valid disclaimer), it is slowing down.  So ideally we only invoke it once upon "Next". This closes this [issue](https://www.notion.so/curvelabs/339a249185e14ddcbd731e5ba07c46c7?v=f7cba822bb9043a5a87ab80b5e2b40ef&p=522c9394058849cf803d7de586940978)


src/newSeed/baseStage.ts
```diff
@@ -31,6 +31,7 @@ export abstract class BaseStage {
   protected maxStage: number;
   protected stageStates: Array<IStageState>;
   protected wizardState: IWizardState;
+  protected nextButtonClicked: boolean;
```
Peartes, line 34: Here I am keeping track if the next button was clicked, this way I can tell that the validate input function has been called

src/newSeed/baseStage.ts
```diff
@@ -43,13 +44,16 @@ export abstract class BaseStage {
   }
 
   activate(_params: unknown, routeConfig: RouteConfig): void {
+    this.nextButtonClicked = false;
```
Peartes, line 47: Set the next button clicked to false every time we navigate to any stage

src/newSeed/baseStage.ts
```diff
@@ -43,13 +44,16 @@ export abstract class BaseStage {
...
     Object.assign(this, routeConfig.settings);
   }
 
   async detached(): Promise<void> {
-    const message = await this.validateInputs();
-    if (!message) {
-      this.persistData();
+    if (!this.nextButtonClicked) {
```
Peartes, line 52: Only call the validate input function if the next button has not been clicked, this means that the navigation away was from the side-bar

src/newSeed/stage3.ts
```diff
@@ -40,6 +40,8 @@ export class Stage3 extends BaseStage {
     if (message) {
       this.validationError(message);
     } else {
+      this.persistData();
```
Peartes, line 43: Some stages have a function persistData() that saves some information. This persisData() is previously only called in the detached function of each stage but since this function in the detached function is called only if the next button was clicked, I made sure to call this function upon validation using the next button before proceeding

PR #122, opened 2021-07-15
made the left-side corneres round too

giladt:
In order to maintain the radius correctly in both mobile and desktop views, I had to take out the `right-margin` from the 'buttons' `contribute` and `launch` and use `gap` instead.

Closes this [Issue](https://www.notion.so/curvelabs/Round-the-Corners-of-Contribute-Apply-for-A-Launch-caaf16503b8848f793b351045e8a45b7)

### DIFF
--- a/src/home/home.scss
+++ b/src/home/home.scss
@@ -51,25 +51,29 @@
         justify-content: space-between;
         cursor: pointer;
         width: 710px;
-
+        gap: 24px;
+        .buttonContainer {
+            display: flex;
+            border-radius: 6px;
+            overflow: hidden;
+        }
         .contribute,
         .launch {
           border-left-width: 4px;
           border-left-style: solid;
           border-image: linear-gradient(to bottom, #ff497a, #a258a7) 1 100%;
-          border-top-right-radius: 6px;
-          border-bottom-right-radius: 6px;
+          border-radius: 6px;
           background-color: $BG02;
           width: 335px;
 
           padding: 18px;
-          padding-right: 24px;
 
           .title {
             display: flex;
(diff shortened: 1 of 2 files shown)


src/home/home.scss
```diff
@@ -51,25 +51,29 @@
         justify-content: space-between;
         cursor: pointer;
         width: 710px;
-
+        gap: 24px;
```
dkent600, line 54: We can't use "gap" or "row-gap" or "column-gap" with flex.  (Doesn't work reliably with iOS).  But we can use them with grid, which is why our best practice, except in trivial cases, is to use grid, not flex.

But it looks great to me.  Switch to `grid` and this should be acceptable.

dkent600 (approved):

dkent600 (changes requested):
Supposed to be centered

image

src/home/home.scss
```diff
@@ -47,29 +47,34 @@
...
 
       .buttons {
-        display: flex;
+        display: grid;
+        grid-auto-flow: column;
         justify-content: space-between;
         cursor: pointer;
         width: 710px;
-
+        grid-gap: 24px;
```
dkent600, line 55: this affects both column and row.  If the intent here is to only set the column (which I think it is), then please use column-gap.

dkent600 (changes requested):

src/home/home.scss
giladt: I use grid-gap here intentionally. It serves the gap for both desktop and mobile view (when it is wrapped).

dkent600 (approved):

PR #125, opened 2021-07-22
Disclaimer decline button text is not visible in safari

giladt:
Apparently the `-webkit-text-fill-color: transparent;` is what’s failing in Safari, but only when used with a `button` tag. Once I change the button into a div tag (and set it to `inline-block`) it functions and looks correctly on chrome/safari/opera/edge/brave/firefox browsers.

I found this [issue](https://www.notion.so/curvelabs/Disclaimer-DECLINE-button-text-is-not-visible-in-Safari-16ef63b72fa0480296dedd79744cd998) affecting both the **disclaimer** and the **stage4** page of the wizard.

### DIFF
--- a/src/home/home.scss
+++ b/src/home/home.scss
@@ -323,18 +323,3 @@
     }
   }
 }
-
-@media screen and (max-width: 420px) {
-  .page.home {
-    .section {
-      &.top {
-        .buttons {
-          .contribute,
-          .launch {
-            width: unset;
-          }
-        }
-      }
-    }
-  }
-}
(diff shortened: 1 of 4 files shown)


src/home/home.scss
```diff
@@ -323,18 +323,3 @@
     }
   }
 }
-
-@media screen and (max-width: 420px) {
```
dkent600, line 327: This should be there, no?  Was part of a previous PR?

dkent600 (changes requested):

src/home/home.scss
giladt: No, this part is from the previous PR. I'm not sure why it has been recommitted, but it should not be included in the code.

dkent600: Can you remove it then?

dkent600 (changes requested):

giladt:
Created a new clean PR #126

PR #128, opened 2021-07-27
fix disclaimer accept button too wide

giladt:
Fixed the accept button to have the same width as the decline button.

I have returned the accept `button` to be a native `button` element, as this is the most simple way to maintain the 'disabled' state that was lost too when converted into `div`.

I've added the two declarations:
```
display: inline-block;
margin: 5px 0 5px 5px;
```
they defined for the native `button` and should have been translated to the `div`-button too.

### DIFF
--- a/src/styles/styles.scss
+++ b/src/styles/styles.scss
@@ -139,8 +139,10 @@ html {
 
     .button1,
     .button2 {
+      display: inline-block;
       border-radius: 50px;
       border: none;
+      margin: 5px 0 5px 5px;
       padding: 14px 28px;
       outline: none;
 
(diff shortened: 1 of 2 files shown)


src/styles/styles.scss
```diff
@@ -139,8 +139,10 @@ html {
 
     .button1,
     .button2 {
+      display: inline-block;
       border-radius: 50px;
       border: none;
+      margin: 5px 0 5px 5px;
```
dkent600, line 145: why do we need this?  Is this the browser-default for button tags?

dkent600 (changes requested):

src/styles/styles.scss
giladt: This is browser default for button tags which I needed to apply for the `div` buttons too.

dkent600 (approved):

PR #129, opened 2021-07-27
Hamburger menu animation

giladt:
Added animation to the menu button in mobile view. [issue](https://www.notion.so/curvelabs/Add-Animation-to-the-Hamburger-menu-on-Mobile-3bc1e9a355fc412987e8d8518a8f6468)

humburger menu animation

### DIFF
--- a/src/app.html
+++ b/src/app.html
@@ -15,12 +15,14 @@
         <div class="headerConnectButton">
           <connectbutton></connectbutton>
         </div>
+        <!--Begin Hamburger Icon -->
         <div class="mobilemenuButton ${showingMobileMenu ? 'showingMobileMenu' : ''}" click.delegate="toggleMobileMenu()">
           <span></span>
           <span></span>
           <span></span>
           <span></span>
         </div>
+        <!-- End Hamburger Icon -->
       </div>
       <div class="mainContent">
         <router-view></router-view>
--- a/src/app.html
+++ b/src/app.html
@@ -3,7 +3,7 @@
   <require from="./headerButtons/headerButtons"></require>
   <banner></banner>
   <modalscreen on-off.to-view="onOff" message.to-view="modalMessage">
-    <div class="appWrapper  ${showingMobileMenu ? 'showingMobileMenu' : ''}">
+    <div class="appWrapper ${showingMobileMenu ? 'showingMobileMenu' : ''}">
       <div class="logo">
         <a href="home"><img src="/primeLaunchLogo.svg" /></a>
       </div>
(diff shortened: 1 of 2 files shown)


src/app.html
```diff
@@ -15,25 +15,32 @@
...
-      <div class="mobileMenuContainer ${showingMobileMenu ? 'showingMobileMenu' : ''}">
-        <div class="menu">
-          <networkfeedback></networkfeedback>
-          <div class="mobileConnectButton">
-            <connectbutton hide-balances></connectbutton>
-          </div>
-          <div class="menuitems">
-            <header-buttons vertical on-navigate.call="onNavigate()"></header-buttons>
-          </div>
+        <div class="mobilemenuButton ${showingMobileMenu ? 'showingMobileMenu' : ''}" click.delegate="toggleMobileMenu()">
```
dkent600, line 18: whaaa?  Why is this needed?  If absolutely necessary, please add a comment in the code.

src/app.scss
```diff
@@ -66,8 +64,13 @@
...
   }
 
+  .appHeader, .logo {
+    top: unset;
+    z-index: 998;
+  }
+
   .mobileMenuContainer {
-    z-index: 998; // stay below modal screen
+    z-index: 997; // stay below modal screen
```
dkent600, line 73: just curious:  why decrease by one?

dkent600 (changes requested):
@giladt There are a ton of changes here that are impossible to review.  Can you please, at this point where we absolutely must limit the number of code changes, and must be carefully reviewing all code changes, minimize the number of code changes and resist the temptation to reorganize code?  I understand you moved some HTML around, which perhaps necessitated a matching reorg of CSS.  So just sayin'...

src/app.html
giladt: Here I've added a new hamburger icon element, which is easier to animate into an `X` when the menu is open.

src/app.html
```diff
@@ -15,25 +15,32 @@
...
       </div>
-      <div show.to-view="!showingMobileMenu" class="mainContent">
+      <div class="mainContent">
         <router-view></router-view>
       </div>
-      <div show.to-view="!showingMobileMenu" class="appFooter">
+      <div class="appFooter">
         <footer></footer>
       </div>
+      <div if.bind="showingMobileMenu" class="animateMobileMenuBackground au-animate">
```
giladt, line 31: I had to move this `mobileMenuContainer` out from the grid, to the parent element in order to animate it on top of the page.

src/app.scss
giladt: I guess it was a left over from some experiment I've done. Changed it back.

dkent600 (approved):

PR #131, opened 2021-08-04
Fix scrollable launch summary

Peartes:
Upon screen sizes smaller than 360px (which is the min-width set on the launch summary container) the launch summary starts to scroll. This commit fixes the responsiveness on such screen sizes.


dkent600 (approved):

dkent600:
@Peartes Before approving I changed the use of "vw" to use %, consistent with the rest of the CSS in that scss file.  Less complexity/risk that way.  Let me know if you see any problem with that.  In the meantime, I'll merge this.

PR #133, opened 2021-08-13
Select package page

Peartes:
Bringing back the select a package page with some CSS cleanup.

- [x]  Keep only the Basic Package and center it
- [x]  It should say: *Use the open-source technology created by PrimeDAO & Balancer to host your own seed or LBP*
- [x]  Can you please bring back this page as given below?
- [x]  Please make sure we aren't using em in the css
- [x]  CSS is properly hierarchically structured
- [x]  if using flex we're going it correctly
- [x]  Remove Prime module completely
- [x]  Make sure that the launch page's BG is also displayed on the smaller viewport width

image


src/selectPackage/selectPackage.html
```diff
@@ -1,32 +1,40 @@
 <template>
-  <require from="./packageContainer.html"></require>
   <div class="page animated-page au-animate selectPackage">
     <div class="banner">
       <div class="heading heading1">Select your package</div>
       <div class="subtext">
         <div>
-          PrimeDAO can help you host a succesfull permissionless token launch.
+          PrimeDAO can help you host a succesfull permissionless token launch. Select the package that firts best for your project.
```
dkent600, line 7: firts => works

src/home/home.html
```diff
@@ -20,7 +20,7 @@
           </div>
         </div>
         <div class="buttonContainer">
-          <div class="launch" click.delegate="navigate('launch')">
+          <div class="launch" click.delegate="navigate('selectPackage')">
```
dkent600, line 23: Actually, this change should be reverted.  It is the launch page link that needs to go to selectPackage.

dkent600 (changes requested):

src/home/home.html
Peartes: Yes true, reverting

src/selectPackage/selectPackage.html
```diff
@@ -1,32 +1,40 @@
 <template>
-  <require from="./packageContainer.html"></require>
   <div class="page animated-page au-animate selectPackage">
     <div class="banner">
       <div class="heading heading1">Select your package</div>
       <div class="subtext">
         <div>
-          PrimeDAO can help you host a succesfull permissionless token launch.
+          PrimeDAO can help you host a succesfull permissionless token launch. Select the package that works best for your project.
```
dkent600, line 7: Sorry I missed this the first review,  "succesfull" should be "successful"

dkent600 (changes requested):
Per the mock, this page should have a background image .  Maybe is the same as the Launch page.

dkent600 (approved):

PR #136, opened 2021-08-18
Documentation navigation

Peartes:
This commit adds the following feature:

- [x] Add next and previous buttons to the end of the page
- [x] The button (or link) text should be the title of the document.
- [x] No containing boxes
- [x] No need to say "Next" or "Previous". Left and right arrows are ok.
- [x]  Only show the next and previous arrow if not the beginning or the end doc

Fixes this [issue](https://www.notion.so/curvelabs/339a249185e14ddcbd731e5ba07c46c7?v=f7cba822bb9043a5a87ab80b5e2b40ef&p=19692e8c1f4e487ab044965f6bb215fa)

Screenshot of appearance below
image


src/documentation/documentation.ts
```diff
@@ -1,12 +1,32 @@
...
+    }
+  }
+
+  @computedFrom("router.currentInstruction")
+  get previousDocTitle(): string {
+    const docNumber = this.router.currentInstruction.config.settings.docNumber;
+    if (docNumber > 1) {
+      return this.router.routes[docNumber - 1].title;
+    } else {
+      return this.router.routes[docNumber].title;
```
dkent600, line 26: remove this

dkent600 (changes requested):
1. the links should be the gradient, like everywhere else in the app.   (Note that red usually signifies an error.)  Possibly the arrows should have their own color, I'm not sure.
2. the back and next buttons should not be visible when at the beginning or end

dkent600 (changes requested):
The place where the blank buttons are still show a "hot" cursor.  Should be a regular pointer.

And for visible buttons, there is a whole lot of blank space that has a "hot" cursor.

Only the text of the button should be "hot"

dkent600 (approved):

dkent600 (changes requested):
You still haven't displayed the buttons as gradients, like elsewhere in the app

dkent600 (approved):

PR #138, opened 2021-08-20
Public seed fee

Peartes:
This commit addresses this [issue](https://www.notion.so/curvelabs/339a249185e14ddcbd731e5ba07c46c7?v=f7cba822bb9043a5a87ab80b5e2b40ef&p=368198bb2a044d01af10f7673c4411f4) by doing the following:

- [x] Define seedFeed as a single public constant number value in the SeedService.ts file
- [x]  The value should be 0  (yes, 0)
- [x] Use this public constant in places where seedFee is referenced


src/services/SeedService.ts
```diff
@@ -27,6 +27,7 @@ export interface ISeedCreatedEventArgs {
 export class SeedService {
 
   public seeds: Map<Address, Seed>;
+  public static seedFee = 0;
```
dkent600, line 30: lets make this 0.0.  Add a code comment "If the value is ever > 0, then should be a fraction like 0.1 to represent 1%".

src/selectPackage/selectPackage.ts
```diff
@@ -1,4 +1,10 @@
+import { SeedService } from "services/SeedService";
 import "./selectPackage.scss";
 
 export class SelectPackage {
+  private seedFee: number;
+
+  constructor() {
+    this.seedFee = SeedService.seedFee;
```
dkent600, line 8: Should be * 100

dkent600 (changes requested):

src/services/SeedService.ts
```diff
@@ -204,7 +205,7 @@ export class SeedService {
       Date.parse(config.seedDetails.endDate) / 1000,
       [config.seedDetails.vestingPeriod, config.seedDetails.vestingCliff],
       !!config.seedDetails.whitelist,
-      toWei(".02"), // 2%
+      toWei(SeedService.seedFee), // 2%
```
dkent600, line 208: delete the code comment, which is incorrect and not needed

dkent600 (changes requested):

dkent600 (approved):

Issue #193, opened 2021-09-10
Select Package: Start registration button should act same as other buttons when hovered over


BartuCulha:
please remove the decotarion and the blue cover when hovered over

![Select-Your-Package-PrimeLaunch (1).png](https://images.zenhubusercontent.com/609a9f5f0a40bdfde66fe338/d48e783d-0dd0-4a20-a7fa-b7ba42b88c3e)

PR #213, opened 2021-09-15
converted link into a button

giladt:
This PR fixes the button behavior on the `Select Package` page- Should be a button and not a link.
Closes Issue #193.

### DIFF
--- a/src/selectPackage/selectPackage.html
+++ b/src/selectPackage/selectPackage.html
@@ -30,7 +30,7 @@
               If successful
             </div>
           </div>
-          <a class="button button1" href="/newSeed">START REGISTRATION</a>
+          <a class="button1" href="/newSeed">START REGISTRATION</a>
           <div class="learnMore">
             <a>LEARN MORE</a>
           </div>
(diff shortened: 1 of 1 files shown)


src/selectPackage/selectPackage.html
```diff
@@ -30,9 +30,9 @@
               If successful
             </div>
           </div>
-          <div class="button button1">
-            <a href="/newSeed">START REGISTRATION</a>
-          </div>
+          <button click.delegate="navigate('newSeed')" class="button button1">
```
dkent600, line 33: image

dkent600: I think the best way to fix this is to leave it as an anchor button but apply the .button1 style to it.

note that a ".button" style doesn't even exist, which is by design.

dkent600 (changes requested):

src/selectPackage/selectPackage.html
```diff
@@ -30,9 +30,7 @@
               If successful
             </div>
           </div>
-          <div class="button button1">
-            <a href="/newSeed">START REGISTRATION</a>
-          </div>
+          <a class="button button1" href="/newSeed">START REGISTRATION</a>
```
dkent600, line 33: remove "button" class

dkent600 (changes requested):

dkent600 (approved):

Issue #191, opened 2021-09-09
Align Tic and Cross Icons in Select Package page


BartuCulha:

BartuCulha:
Please align the tics and cross icons to the middle of the circle
![Select Your Pack](https://user-images.githubusercontent.com/24368665/132711413-a6687e38-abfe-4723-bd1c-90eae6eb32f4.png)

dkent600:
@renc-curvelabs This doesn't have a priority set

renc-curvelabs:
Yes - it is on purpose 
I would like to see how it looks with the new copy first

dkent600:
Then should remain in either in spec or needs triage

giladt:
This have been already done in #185

Issue #185, opened 2021-09-03
Select Package Copy


renc-curvelabs:

renc-curvelabs:
![image](https://user-images.githubusercontent.com/79934562/131998350-c903ed83-b412-452c-8107-cbd649408a82.png)

Hey Ez, we need a new copy for this page as now we are going with providing no services (marketing should also be removed). Can you please provide one? 

Thanks!

EzraWeller:
Can you give me a clearer idea of what you have in mind? It looks like there is only one package available -- why are we even offering people to "select a package"? What is this page for exactly if there's only one package?

renc-curvelabs:
Hey Ez, 

Here are some notes from our discussions with Luuk about this page:

- We will be offering only 1 package at this time due to legal regulations we still have to figure out. This might change in the future. 
- This package will be free and will not include Marketing. So we can get rid of that bulletpoint all together. 
- We want to display this page to make an emphasis that they are getting this for free (For now). As people do like free things and us asking for no fees is a difference maker at this point. 
- Feel free to change the copy however you seem fit (also including the title, if "selecting" a package does not make sense)

EzraWeller:
Okay, sounds good. So "Support" and "Hosted" are still being offered, yes? Can I get more detail on what those are? To me it looks like the main problem is that it isn't clear what those mean. 

The implication of keeping the page as "Select your package" with just one package is that there will be more packages in the future -- is that the plan still? If that's uncertain I would suggest getting rid of this page and adding the benefits and price to the previous page under "Seed."

renc-curvelabs:
Support: They will have access to a support team which they can forward feedback towards. They will basically have support agent or agents
Hosted: Their sale will be displayed in the platform 

We will keep this page as we will have more packages in the long term

renc-curvelabs:
Hey Ez - coming back to this 
Can we finalize something today? Thanks!

EzraWeller:
Something like this? Has "liquid launch" changed names to "LBP"? "LBP" doesn't seem great, but if that's what we're using atm, cool.

---
# Select your package
More packages will be available in the future.

## Basic
Seed or LBP launch
- [x] Featured Listing -- your launch will be featured in Prime Launch's curated list of token launches 
- [x] Personal Support -- Prime DAO will assign 1-2 community representatives to guide you through interacting with the platform and the community

### 0% fee
for approved launches

### Start Registration
### Learn More

---

[...]

PR #214, opened 2021-09-15
Select package copy

giladt:
1. Applied the new copy from @EzraWeller. I have split the description into its own div and gave it a little styling. Closing issue #185.
2. Centered the `Check-Icons` closing issue #191 .
3. Fixed the green color of $AlertGreen. Applied this color to the `Check-Icons`

### DIFF
--- a/src/selectPackage/selectPackage.html
+++ b/src/selectPackage/selectPackage.html
@@ -7,7 +7,7 @@
       <div class="heading heading1">Select your package</div>
       <div class="subtext">
         <div>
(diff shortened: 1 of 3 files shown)


dkent600 (changes requested):
Can this gap be a little smaller?  Seems too large.

image

dkent600:
Title should not be "Select your package".  If that is not spec'd, it should be.

dkent600 (approved):

PR #218, opened 2021-09-16
Refactor/dropdown animation

giladt:
Added smoother animation for the border-radius interface of a closed/open dropdown.


dkent600 (changes requested):
fix merge conflicts

dkent600 (changes requested):
The animation tends to be very easily interrupted by other processes running.  For example, show to stage 4 and select a funding token.  On my machine the slide up is interrupted halfway through (I think usually when selecting a token for the first time).  Is it possible to give the animation higher priority?

dkent600 (approved):

PR #222, opened 2021-09-27
Style/flex whitespace and refactor navbar

giladt:
1. Refactored the whitespace behaviour to adopt better to wide screens
   - No fixed 160px margins. Instead fixed main container width up to 1440px, then flexible with margins 50px (to align with the logo), and 20px on smaller screens
   - fixed the moonshot position to align with new behavior.
   - fixed the launch page background image to align correctly to the right (here we could debate if it shouldn't be centered with the main content)
   - nomiss e-mail form brakes vertically on smaller screens

2. Made adjustments to the navbar and footer
   - Fixed non active icons colors
   - Fixed height and margins 
   - Fixed footer width brighter then the view width
   - Set the page min-height to view height. This way the footer is always at the bottom of the screen.


dkent600:
Closing because this is too much, need to make smaller PRs and be in sync with the kanban.

PR #224, opened 2021-09-27
add loading state to token-select

giladt:
Added Prefill and loading state to the `token-select` element.
Small fix to the `dropdown` height.

### DIFF
--- a/src/resources/elements/dropdown/dropdown.scss
+++ b/src/resources/elements/dropdown/dropdown.scss
@@ -34,7 +34,6 @@
     cursor: pointer;
     margin: 11px;
     margin-right: 0;
-    height: 21px;
 
     .itemContent {
       margin-right: 8px;
--- a/src/resources/elements/dropdown/dropdown.scss
+++ b/src/resources/elements/dropdown/dropdown.scss
@@ -34,6 +34,7 @@
     cursor: pointer;
     margin: 11px;
     margin-right: 0;
+    height: 21px;
 
     .itemContent {
       margin-right: 8px;
--- a/src/resources/elements/tokenSelect/tokenSelect.scss
+++ b/src/resources/elements/tokenSelect/tokenSelect.scss
@@ -9,10 +9,18 @@
     align-items: center;
     height: 23px;
 
-    img {
+    img, i {
       @include tokenIcon;
       position: relative;
+      text-align: center
+    }
+
+    img {
       top: 2px;
+    }
+
+    i {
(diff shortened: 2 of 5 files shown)


src/resources/elements/tokenSelect/tokenSelect.scss
```diff
@@ -9,12 +9,20 @@
...
       position: relative;
+      text-align: center
+    }
+
+    img {
       top: 2px;
     }
 
+    i {
+      margin-top: 2px;
```
dkent600, line 23: why aren't you using top or bottom, like `img` does, since the position is relative?

src/resources/elements/tokenSelect/tokenSelect.ts
```diff
@@ -30,6 +30,7 @@ export class TokenSelect {
 
   async attached(): Promise<void> {
     if (!this.tokenInfos) {
+      this.tokenInfos = Array(this.tokenAddresses.length).fill({} as ITokenInfo);
```
dkent600, line 33: since we are waiting for the entire list, not for individual items, I might have put the spinner in the placeholder instead of every item.  Would that make sense?  Could be simpler to code and better UX.

src/resources/elements/dropdown/dropdown.scss
```diff
@@ -34,6 +34,7 @@
     cursor: pointer;
     margin: 11px;
     margin-right: 0;
+    height: 21px;
```
dkent600, line 37: Why?  this makes me nervous because it has potential wide effect beyond what you are trying to achieve here.

dkent600 (changes requested):

src/resources/elements/tokenSelect/tokenSelect.ts
giladt: I've simplified the code. commit a19ce14

src/resources/elements/dropdown/dropdown.scss
giladt: reverted in commit 3bb80d3. Instead I've changed the tokenInfo height. This is in order to keep the height from changing and dragging the whole view below up or down, in case there's multiple icon sizes.

src/resources/elements/tokenSelect/tokenSelect.scss
giladt: reverted in commit 3bb80d3 as it is not necessary anymore.

dkent600 (approved):

PR #254, opened 2021-10-04
nicer spinny icon for token select dropdown

### LINKED ISSUE #245: nicer spinny icon for token select dropdown
Can we put a little more space between the spinny icon and "Select", and give the icon some color, maybe $Secondary03-dev?

image

giladt:
See #245 

Looking at common UI's, might make sense to refactor the drop down to have a `loading` state, and replace the spinner with the `arrow` icon on the right, while loading.
Currently it's impossible to implement since the loading state applies only to the `select-token` element.

Suggested as an optional change...


dkent600:
> Looking at common UI's, might make sense to refactor the drop down to have a loading state, 

I thought about that from the beginning.  Ultimately I fell on the side of letting the dropdown consumer control the state and appearance when waiting.  For now is good enough

dkent600 (approved):

PR #256, opened 2021-10-04
Fix/mobile geoblock

### LINKED ISSUE #244: Registration Wizard: Geoblock button needs adjustments

giladt:
See #244 

1. On mobile labels have 5px margin. @dkent600 is it necessary? For this issue I've removed the margin only from the `geoBlock` label, but if not needed, it can be removed globally- what caused this issue in the first place.
2. I've also added a small fix to the left-position of the toggle-handle (applies all screen sizes).

### DIFF
--- a/src/newLaunch/baseStage.scss
+++ b/src/newLaunch/baseStage.scss
@@ -709,6 +709,10 @@
       line-height: 20px;
       margin-bottom: 5px;
     }
+
+    #geoBlock+label {
+      margin-bottom: unset;
+    }
     .summaryContainer {
       margin-bottom: 30px;
       word-break: break-word;
--- a/src/newLaunch/baseStage.scss
+++ b/src/newLaunch/baseStage.scss
@@ -574,7 +574,7 @@
           width: 22px;
           height: 22px;
           position: absolute;
-          left: 2px;
+          left: 4px;
           background: white;
           border-radius: 50%;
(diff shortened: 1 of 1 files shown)


dkent600:
> On mobile labels have 5px margin. @dkent600 is it necessary? For this issue I've removed the margin only from the geoBlock label, but if not needed, it can be removed globally- what caused this issue in the first place.

@giladt I don't know, need you to figure that out and do the best thing. on the one hand, the simpler solution is lower risk, but would be nice to get rid of this margin setting if is unneeded.

giladt:
> > On mobile labels have 5px margin. @dkent600 is it necessary? For this issue, I've removed the margin only from the geoBlock label, but if not needed, it can be removed globally- what caused this issue in the first place.
> 
> @giladt I don't know, need you to figure that out and do the best thing. on the one hand, the simpler solution is lower risk but would be nice to get rid of this margin setting if is unneeded.

I had a deeper look at the effects of this `margin-bottom` declaration. Removing it will affect all labels with no question marks in the wizard. So removing it now might have a too risky impact. 

Eventually, I'd suggest a complete clean-up of the label's styling in the wizard, as we have multiple unnecessary margin definitions for every label version!

dkent600:
@giladt Totally agree about cleaning up the CSS in baseStage.scss.

dkent600 (approved):

PR #278, opened 2021-10-07
Fixed missing gap in labels

giladt:
Sets a default box-sizing in order to prevent misbehaviour.

I couldn't observe the issue in local, so this fix is more of a guess. 
Adding this fix in the testing site console tough, have solved the issue.


src/styles/styles.scss
```diff
@@ -32,6 +32,7 @@ html {
     min-height: 100vh;
     padding-left: $BodyPaddingLeftRight;
     padding-right: $BodyPaddingLeftRight;
+    box-sizing: content-box;
```
dkent600, line 35: ouch, this is a global change.  Have you gone though the entire app to confirm this doesn't alter anything else?  Be sure to look at the busy/spinny background and how it fits the screen.  everything!

But I think you should make this change apply specifically to the case referenced in the ticket.

PR #279, opened 2021-10-07
Removed unnecessary margin from button

testing-life:
Fixed path typo


dkent600 (changes requested):
Sorry, I approved and merged this too hastily.  This is a change that will affect every button in the application.  I can't approve as-is.  We need to reevaluate the change and submit a new PR on it.

I reverted the merge of this PR

PR #280, opened 2021-10-07
Added LBP Token Price Service

giladt:
Added the calculations from @fabianschu to the repo.
Testing needed.


dkent600:
@giladt maybe could be a good case for a typescript-only unit test?

src/services/LbpProjectTokenPriceService.ts
```diff
@@ -0,0 +1,144 @@
+import { autoinject } from "aurelia-framework";
+import * as moment from "moment-timezone";
+import Moment = moment.Moment;
+
+@autoinject
+export class LbpProjectTokenPriceService {
+
+  maxSupply: number;
```
dkent600, line 8: in services, be explicit about what is public vs private

src/services/LbpProjectTokenPriceService.ts
```diff
@@ -0,0 +1,144 @@
...
+  startDate: string;
+  endDate: string;
+  roundedStartDate: Moment;
+  roundedEndDate: Moment;
+  lbpDurationInHours: number;
+  pricePerFundingToken: number;
+
+
+  constructor(
+    maxSupply: number,
```
dkent600, line 22: you can specify these constructor params as private and they will be automatically initialized to the instance variables of the same name.

src/services/LbpProjectTokenPriceService.ts
```diff
@@ -0,0 +1,144 @@
...
+    projectTokenInitialLbpAmount: number,
```
dkent600, line 23: I'm pretty sure some of these are BigNumbers, not numbers

src/services/LbpProjectTokenPriceService.ts
```diff
@@ -0,0 +1,144 @@
...
+import * as moment from "moment-timezone";
+import Moment = moment.Moment;
+
+@autoinject
+export class LbpProjectTokenPriceService {
+
+  maxSupply: number;
+  projectTokenInitialLbpAmount: number;
+  fundingTokenInitialLbpAmount: number;
+  startWeightProjectToken: number;
```
dkent600, line 11: why specify "ProjectToken"?  Is there any other kind of weight?  Feel free to document...

src/services/LbpProjectTokenPriceService.ts
```diff
@@ -0,0 +1,144 @@
...
+  endWeightProjectToken: number;
+  startDate: string;
+  endDate: string;
+  roundedStartDate: Moment;
+  roundedEndDate: Moment;
+  lbpDurationInHours: number;
```
dkent600, line 17: can't we compute this from start and end date?

src/services/LbpProjectTokenPriceService.ts
```diff
@@ -0,0 +1,144 @@
...
+export class LbpProjectTokenPriceService {
+
+  maxSupply: number;
+  projectTokenInitialLbpAmount: number;
+  fundingTokenInitialLbpAmount: number;
+  startWeightProjectToken: number;
+  endWeightProjectToken: number;
+  startDate: string;
+  endDate: string;
+  roundedStartDate: Moment;
```
dkent600, line 15: rounder how?  can we do this rounding in the service rather than forcing the caller to do it?

src/services/LbpProjectTokenPriceService.ts
dkent600, line 8: specify which token

src/services/LbpProjectTokenPriceService.ts
```diff
@@ -0,0 +1,144 @@
...
+    return Math.round(projectTokenMcap);
+  }
+
+  getPriceAtWeight(
+    projectTokenInPool: number,
+    fundingTokenInPool: number,
+    projectTokenWeight: number,
+    pricePerFundingToken: number,
+  ): number {
+    const fundingTokenValue = fundingTokenInPool * pricePerFundingToken;
```
dkent600, line 68: I think you should be using BIgNumber methods.  You will need to use our BigNumberService (BigNumber.js) because the usual ethers BigNumber doesn't do division.

You may need to account for the # of decimals of each token.  You can talk to the contract guys about that.

src/services/LbpProjectTokenPriceService.ts
giladt: I took it from Fabi's code, and thought the same too at the begining. But I guess it is more prominent to say that it's the projectToken weight in the pair - (80% PT) <-> (20% FT)

src/services/LbpProjectTokenPriceService.ts
```diff
@@ -0,0 +1,152 @@
+import { autoinject } from "aurelia-framework";
+import * as moment from "moment-timezone";
+import BigNumber from "bignumber.js";
```
dkent600, line 3: best to import BigNumber from services/BigNumberService

PR #286, opened 2021-10-07
Disclaimer input spacing

testing-life:


src/resources/dialogs/dialogs.scss
```diff
@@ -83,6 +82,9 @@ body > ux-dialog-container {
     }
 
     &.disclaimer {
+      .marginRight10 {
+        margin-inline-end: 10px;
```
dkent600, line 86: Lets just use margin-right here.  We only support english, don't know and aren't testing (and don't want to test) the implications of using margin-inline-end on this element for all cases, and don't want to introduce this kind of thing into the app without a clear plan/intention and being consistent with it across the app.

src/resources/dialogs/dialogs.scss
```diff
@@ -83,6 +82,9 @@ body > ux-dialog-container {
     }
 
     &.disclaimer {
+      .marginRight10 {
```
dkent600, line 85: I find your thoughts on this are valuable! But for the reasons I gave, and to make this code consistent with the rest of the app, and to make it easy to write and locate the styles that apply to this specific element, lets call this class "disclaimerCheckbox", or any name that adequately identifies this element and is properly scoped to only apply to this element.  Delete `marginRight10 `.

This project is just not positioned at this time to mix in a significantly different, alternative means of organizing classes, and the pros and cons have not been discussed as a team.

testing-life: due to the time constraint, fair enough, but i'd appreciate if you could point me to some resource detailing our naming methodology, so that i could fully understand it :) i'm guessing our doc added several days ago is just a list of points to discuss after the november deadline, right ?

dkent600 (approved):

PR #318, opened 2021-10-08
Fix: Added fix for `Retrieve Project Token`.

codebuster22:
`Retrieve Project Token` feature was displayed for `Incomplete` Seed.

### DIFF
--- a/src/admin/seeds/dashboard/dashboard.html
+++ b/src/admin/seeds/dashboard/dashboard.html
@@ -78,12 +78,12 @@
                 </div>
               </div>
             </div>
-            <div class="featureContainer" if.bind="selectedSeed.projectTokenBalance >= selectedSeed.seedRemainder && (selectedSeed.isClosed || selectedSeed.claimingIsOpen || selectedSeed.incomplete)">
+            <div class="featureContainer" if.bind="selectedSeed.projectTokenBalance >= selectedSeed.seedRemainder && (selectedSeed.isClosed || selectedSeed.maximumReached || selectedSeed.isDead)">
               <div class="featureTitle">
                 Retrieve Project Tokens
               </div>
               <div class="featureDescription">
-                ${selectedSeed.projectTokenBalance.sub(selectedSeed.seedRemainder.add(selectedSeed.feeRemainder)) | ethwei:selectedSeed.projectTokenInfo.decimals} ${selectedSeed.projectTokenInfo.name} (${selectedSeed.projectTokenInfo.symbol}) <img class="tokenImage" src.to-view="selectedSeed.projectTokenInfo.logoURI"></img> will be transferred to the address you provide below
+                ${retrievableProjectTokenAmount() | ethwei:selectedSeed.projectTokenInfo.decimals} ${selectedSeed.projectTokenInfo.name} (${selectedSeed.projectTokenInfo.symbol}) <img class="tokenImage" src.to-view="selectedSeed.projectTokenInfo.logoURI"></img> will be transferred to the address you provide below
               </div>
               <div>
                 <input id="receiverAddress" value.bind="receiverAddress" placeholder="0x6C4ef..." />
--- a/src/admin/seeds/dashboard/dashboard.html
+++ b/src/admin/seeds/dashboard/dashboard.html
@@ -78,7 +78,7 @@
                 </div>
               </div>
             </div>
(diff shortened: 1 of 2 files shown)


src/admin/seeds/dashboard/dashboard.html
```diff
@@ -78,12 +78,12 @@
                 </div>
               </div>
             </div>
-            <div class="featureContainer" if.bind="selectedSeed.projectTokenBalance >= selectedSeed.seedRemainder && (selectedSeed.isClosed || selectedSeed.claimingIsOpen)">
+            <div class="featureContainer" if.bind="selectedSeed.projectTokenBalance >= selectedSeed.seedRemainder && (selectedSeed.isClosed || selectedSeed.maximumReached || selectedSeed.isDead)">
```
dkent600, line 81: Why are you not just using `seed.claimingIsOpen`?

src/admin/seeds/dashboard/dashboard.html
```diff
@@ -78,12 +78,12 @@
...
               <div class="featureTitle">
                 Retrieve Project Tokens
               </div>
               <div class="featureDescription">
-                ${selectedSeed.projectTokenBalance.sub(selectedSeed.seedRemainder.add(selectedSeed.feeRemainder)) | ethwei:selectedSeed.projectTokenInfo.decimals} ${selectedSeed.projectTokenInfo.name} (${selectedSeed.projectTokenInfo.symbol}) <img class="tokenImage" src.to-view="selectedSeed.projectTokenInfo.logoURI"></img> will be transferred to the address you provide below
+                ${retrievableProjectTokenAmount() | ethwei:selectedSeed.projectTokenInfo.decimals} ${selectedSeed.projectTokenInfo.name} (${selectedSeed.projectTokenInfo.symbol}) <img class="tokenImage" src.to-view="selectedSeed.projectTokenInfo.logoURI"></img> will be transferred to the address you provide below
```
dkent600, line 86: have you tested whether this amount will be updated if the selectedSeed changes? If not, then either use a getter property (decorated appropriately with `@computetFrom`), or pass in selectedSeed as an argument.

dkent600 (changes requested):

src/admin/seeds/dashboard/dashboard.html
codebuster22: These are the conditions that are used in contract to check if `retrieveSeedTokens` can be invoked or not.

dkent600: ah, i get it.

src/admin/seeds/dashboard/dashboard.ts
```diff
@@ -8,6 +8,7 @@ import { EventAggregator } from "aurelia-event-aggregator";
 import { DisposableCollection } from "services/DisposableCollection";
 import { EventConfigException } from "services/GeneralEvents";
 import { WhiteListService } from "services/WhiteListService";
+import { BigNumber } from "@ethersproject/providers/node_modules/@ethersproject/bignumber";
```
dkent600, line 11: Should be like in the rest of our project, and not dependent on package path:

`import { BigNumber } from "ethers";`

src/admin/seeds/dashboard/dashboard.html
```diff
@@ -78,12 +78,12 @@
                 </div>
               </div>
             </div>
-            <div class="featureContainer" if.bind="selectedSeed.projectTokenBalance >= selectedSeed.seedRemainder && (selectedSeed.isClosed || selectedSeed.claimingIsOpen)">
+            <div class="featureContainer" if.bind="retrievableProjectTokenAmount != 0 && selectedSeed.projectTokenBalance >= selectedSeed.seedRemainder && (selectedSeed.isClosed || selectedSeed.maximumReached || selectedSeed.isDead)">
```
dkent600, line 81: do we really need this `selectedSeed.projectTokenBalance >= selectedSeed.seedRemainder` if we already are using this `retrievableProjectTokenAmount `

Are you sure this shouldn't be `retrievableProjectTokenAmount.gt(0)`?

codebuster22: yes, I realized now.

src/admin/seeds/dashboard/dashboard.html
codebuster22: this won't be used as if selectedSeed is null, the sections that use this won't be rendered. I have added an if condition, that if selectedSeed is undefined, `retrievableProjectTokenAmount` will return 0.

PR #322, opened 2021-10-08
Refactor/236 eliminate redundant code

giladt:
Current status:

* Moved all Seed/LBP similar stages to shared `stages` folder.

* Stages 4/6/7 are different for each product and reside in their own `stages` folder under the corespondent product folder.

* Updates to file names:
  * seed / lbp:
    * lbpLaunch --> launch
    * lbpConfig --> config

* Updated names:
    Stage3: maxSeedSupply / maxProductTokenSupply -> maxSupply  !! Important - needs update to the contracts or finding a different solution


src/newLaunch/launchConfig.ts
```diff
@@ -19,8 +18,8 @@ export const CategoryNames = [
 ];
 
 export class SocialLinkSpec {
-  public media: string;
-  public url: string;
+  public media: string
```
dkent600, line 21: please run `npm run lint`

src/newLaunch/launchConfig.ts
```diff
@@ -36,15 +35,15 @@ export interface IProjectDetails {
   summary: string,
   proposition: string,
   teamDescription: string,
-  logo: string
+  logo?: string
```
dkent600, line 38: why is this marked optional?  should this actually be deleted?  I can't find where it is used.

src/newLaunch/lbp/config.ts
```diff
@@ -0,0 +1,77 @@
...
+  IProjectDetails,
+  ITokenDetails,
+  IContactDetails,
+} from "newLaunch/launchConfig";
+
+export interface ILbpDetails {
+  fundingTokenAddress: string,
+  amountProjectToken: string,
+  amountFundingToken: string,
+  startDate: string,
```
dkent600, line 13: I propose that this and other props here should be in a ILaunchDetails interface that should be part of the shared config interface.

Seed and Lbp config interfaces I think can override the `launchDetails` property like, for example, `launchDetails: ISeedDetails` where `ISeedDetails implements ILaunchDetails`

src/newLaunch/lbp/stages/stage4.ts
```diff
@@ -108,48 +110,48 @@ export class Stage4 extends BaseStage {
...
-  //   this.setLbpConfigEndDate();
-  //   this.wizardState.lbpStartDate = this.lbpConfig.lbpDetails.startDate;
+  //   this.setlaunchConfigStartDate();
+  //   this.setlaunchConfigEndDate();
+  //   this.wizardState.lbpStartDate = this.launchConfig.lbpDetails.startDate;
   // }
 
   toggleLegalDisclaimer(): void {
-    this.lbpConfig.lbpDetails.legalDisclaimer = !this.lbpConfig.lbpDetails.legalDisclaimer;
+    this.launchConfig.lbpDetails.legalDisclaimer = !this.launchConfig.lbpDetails.legalDisclaimer;
```
dkent600, line 146: following from my suggestion above, this would be `this.launchConfig.launchDetails`

dkent600 (changes requested):

src/newLaunch/launchConfig.ts
giladt: lint did not catch this missing `;` for some reason.

src/newLaunch/lbp/stages/stage4.ts
giladt: implemented in 596f1f0

src/newLaunch/launchConfig.ts
```diff
@@ -66,16 +65,23 @@ export interface IContactDetails {
   remarks: string
 }
 
+export interface ILaunchDetails {
```
dkent600, line 68: Is there any reason the base ILaunchDetails can't also contain fundingTokenInfo, projectTokenInfo and pricePerToken?

dkent600 (changes requested):

src/newLaunch/launchConfig.ts
dkent600: Oops, ok, I think what I should be asking is:  Is there any reason the base ILaunchDetails can't also contain fundingTokenAddress and pricePerToken?

giladt: I've missed the `fundingTokenAddress`. Now fixed.
`pricePerToken` is uniqe to Seed.
The `legalDisclaimer` can also be moved here, but I'll do it within the fix to convert the checkbox into an input field in LBP wizard.

dkent600 (approved):

PR #332, opened 2021-10-11
Vertically aligned inputs with buttons

testing-life:
image

### DIFF
--- a/src/admin/seeds/dashboard/dashboard.html
+++ b/src/admin/seeds/dashboard/dashboard.html
@@ -57,7 +57,7 @@
               <div class="featureDescription">
                 Add an account to the whitelist
               </div>
-              <div>
+              <div class='featureContainer__inputWrapper'>
                 <input id="addressToAdd" value.bind="addressToAdd" placeholder="0x6C4ef..." />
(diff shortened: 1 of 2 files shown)


src/admin/seeds/dashboard/dashboard.html
```diff
@@ -57,7 +57,7 @@
               <div class="featureDescription">
                 Add an account to the whitelist
               </div>
-              <div>
+              <div class='featureContainer__inputWrapper'>
```
dkent600, line 60: Can we be consistent with the class naming here and just call this "inputWrapper"?

testing-life: due to time constraints, fine. :)
for the record though, this is a correct naming approach - it's tightly couple with that element and context, namespace prevents leaking and it's explicit that it refers to this one element, since `inputWrapper` is quite likely to be used for something else, elsewhere.

dkent600: understood.  the reason I am requesting consistency is to avoid needlessly adding another pattern and thus complexity and risk to the app.

PR #339, opened 2021-10-12
Address validation in seed dashboard

testing-life:
For whitelisting, unwhitelisting and retrieving tokens


dkent600 (changes requested):
The validation should not be in Seed, but in the dashboard form before calling the Seed functions.  And the message should specify the input field like:  "Please supply a valid address to whitelist"

testing-life:
> What is the need for the DashboardActions? Seems like unnecessary complication when all we need are three handler functions, one for each action.
> 
> Similarly for AddressValidationMessages when all we need is a single reference to each string, which is only ever used once.

Well, how do I know that it's only used once? How do I know that it will continue being used only once? How do I know that a message has been typed in correctly ? How do I know that more messages won't be needed when requirements/circumstances change? And if they change, do I need to chase through every instance of it being used to change the string? And if I do, how do I know I've changed it correctly everywhere?. If this is driven from one point,  then any changes are easy to add and this increases confidence in code. Your IDE will also give you code hints, again limiting ambiguity. Finally, why do we need 3 handlers, essentially doing very similar things, when we can have just one? Surely, that fulfils the DRY principle, no ? It only achieves what was set out to achieve in the ticket, but it's very open to being extended by another person that may potentially need to touch this functionality.

dkent600 (approved):

dkent600 (approved):

PR #354, opened 2021-10-19
Feature/346 token distribution

testing-life:
A reusable component for token distribution display. Currently without the pie chart, only with a list.
image


src/resources/elements/tokenDistribution/tokenDistribution.html
```diff
@@ -0,0 +1,27 @@
+<template>
+  <section if.bind='data' class="tokenDistribution">
+    <header class="tokenDistributionHeading">
+      <h3 class="heading heading3">Token Distribution</h3>
```
dkent600, line 4: Our code guidelines ask not to use stuff like `header` and `h3`.  Please use div and the closest appropriate style from styles.scss. 

But as long as `header` is known not to introduce formatting, then I wouldn't object to it.

src/seedDashboard/seedDashboard.html
```diff
@@ -192,6 +192,7 @@
         -->
       <div class="section seedSummary">
         <div class="label targetLabel">FUNDING TARGET</div>
+        <token-distribution start-date.bind='seed.metadata.launchDetails.startDate' data.bind='seed.metadata.tokenDetails'></token-distribution>
```
dkent600, line 195: why is this in seedDashboard?

src/resources/elements/tokenDistribution/tokenDistribution.ts
```diff
@@ -0,0 +1,25 @@
+import { ITokenDetails } from "../../../newLaunch/launchConfig";
+import { bindable } from "aurelia-framework";
+import "./tokenDistribution.scss";
+import * as moment from "moment";
```
dkent600, line 4: Can our `DateService` service the purposes here, or can we benefit by adding to it any new and useful functionality found here?

dkent600 (changes requested):

dkent600 (approved):

PR #361, opened 2021-10-21
fix: fix button alignment in stage 4

wantedsystem:


dkent600 (changes requested):
Looks good so far.  Please add a screenshot of the result of your work (desktop and mobile in this  case) and connect the PR to the ticket you are working.  Thanks!

src/newLaunch/seed/stages/stage4.html
```diff
@@ -232,14 +232,16 @@
...
           </div>
-          <input
-            id="admin"
-            value.bind="launchConfig.launchDetails.adminAddress"
-            placeholder="Administrator wallet address" />
-          <template if.to-view="!launchConfig.launchDetails.adminAddress">
-            <div class="button2 small" if.to-view="!ethereumService.defaultAccountAddress" click.delegate="connect()">Connect to get your wallet address</div>
-            <div class="button2 small" else click.delegate="makeMeAdmin()">I will be the administrator</div>
-          </template>
+          <div class="adminInput">
```
giladt, line 235: The `adminInput` class should be applied in the LBP wizard too, since it has the same admin element.
see `src/newLaunch/lbp/stages/stage4.html`

PR #371, opened 2021-10-22
363 - Hydrate current price for projectToken

giladt:
Project Token Price is now available through: `LbpManager.projectTokenInfo.price`

Token Price History is still WIP.


dkent600:
I've already worked this

PR #397, opened 2021-10-29
Trash icon should be on third row

wantedsystem:
Result =>

Capture d’écran 2021-10-29 à 14 09 34

### DIFF
--- a/src/newLaunch/baseStage.scss
+++ b/src/newLaunch/baseStage.scss
@@ -865,19 +865,25 @@
   }
 }
 
-@media (min-width: 240px) and (max-width: 690px) {
+@media screen and (max-width: 690px) {
   .distribRow {
     position: relative;
+    margin-top: 25px;
   }
 
   .deleteLink {
     position: absolute;
     left: 4px;
-    top: 170px;
-    margin-top: 10px;
+    top: 85px;
   }
-  
+
   .addLink {
     margin-top: 50px;
+  }
+}
+
+@media screen and (max-width: 671px) {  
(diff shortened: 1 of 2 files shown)


src/newLaunch/baseStage.scss
```diff
@@ -864,3 +864,26 @@
...
 }
+
+@media screen and (max-width: 690px) {
+  .distribRow {
+    position: relative;
+    margin-top: 25px;
+  }
+
+  .deleteLink {
+    position: absolute;
```
dkent600, line 875: Please do not use `position: absolute`.  This is given in the coding guidelines.

src/newLaunch/baseStage.scss
```diff
@@ -864,3 +864,26 @@
     }
   }
 }
+
+@media screen and (max-width: 690px) {
+  .distribRow {
+    position: relative;
```
dkent600, line 870: why position relative?

src/newLaunch/baseStage.scss
```diff
@@ -864,3 +864,26 @@
...
+    margin-top: 25px;
```
dkent600, line 871: rather than margin-top, can you please configure the flex that is defined by `.distribRow`?  If that is not feasible, switch to grid.

dkent600 (changes requested):
Can you please use a different breakpoint so this doesn't happen?

image

Also add spacing between the trash icon and the last input?  If you need to, make the inputs a little less wide.

image

dkent600 (approved):

PR #400, opened 2021-10-29
Launch Page: Request Information link inside Liquid Launch card needs padding

wantedsystem:
Results => 

Desktop : 
Capture d’écran 2021-10-29 à 17 48 46

Mobile : 
Capture d’écran 2021-10-29 à 17 49 06

### DIFF
--- a/src/launch/launch.scss
+++ b/src/launch/launch.scss
@@ -27,12 +27,16 @@
     flex: 1;
     gap: 24px;
     justify-content: center;
+    
     .title {
       font-size: 36px;
       line-height: 65px;
     }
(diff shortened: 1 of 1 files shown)


dkent600 (changes requested):
Looks bad here.

Please refactor `.launchBoxes` to use grid instead of flex, or at least to use flex properly.  I don't think there is a good way to redeem the existing code; it is always going to be too fragile.

image

dkent600 (approved):
Sorry, I was looking at the wrong branch.  It looks good at all widths, so lets go with this.

PR #411, opened 2021-11-01
launch admin should be required

wantedsystem:
Result =>
Capture d’écran 2021-11-01 à 16 38 02

### DIFF
--- a/src/newLaunch/lbp/stages/stage4.ts
+++ b/src/newLaunch/lbp/stages/stage4.ts
@@ -292,7 +292,7 @@ export class Stage4 extends BaseStage<ILbpConfig> {
       || !(Number.parseInt(endTimes[1]) < 60)) {
       message = "Please enter a valid value for End Time";
     } else if (!Utils.isAddress(this.launchConfig.launchDetails.adminAddress)) {
-      message = "Please enter a valid value for launch administrator";
+      message = "Please enter a valid wallet address for LBP Administrator";
     } else if (this.launchConfig.launchDetails.endWeight > this.launchConfig.launchDetails.startWeight) {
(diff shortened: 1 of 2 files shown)


src/newLaunch/lbp/stages/stage4.ts
```diff
@@ -291,6 +291,8 @@ export class Stage4 extends BaseStage<ILbpConfig> {
     } else if (!(Number.parseInt(endTimes[1]) >= 0)
       || !(Number.parseInt(endTimes[1]) < 60)) {
       message = "Please enter a valid value for End Time";
+    } else if (!Utils.isAddress(this.launchConfig.launchDetails.adminAddress)) {
+      message = "Please enter a valid value for launch administrator";
```
dkent600, line 295: Lets change the text to : "Please enter a valid address for the LBP administrator"

dkent600 (changes requested):
Do we need to fix this for the seed wizard too?

wantedsystem:
> Do we need to fix this for the seed wizard too?

Already exists

src/newLaunch/lbp/stages/stage4.ts
wantedsystem: I changed to : "Please enter a valid wallet address for LBP Administrator", like the seed wizard

dkent600: ok, but for correct english, it should be "the LBP Administrator", in both places ("the Seed Administrator" for seeds).

wantedsystem: Done

dkent600 (approved):

PR #440, opened 2021-11-06
423 - Selected Token Is Clipped By The Dropdown Menu

giladt:
Fixed also the height differences between the dropdown and the input element and the background slipping out of the dropdown border.

Result:
image

### DIFF
--- a/src/lbpDashboard/lbpDashboardForm/lbpDashboardForm.scss
+++ b/src/lbpDashboard/lbpDashboardForm/lbpDashboardForm.scss
@@ -49,7 +49,6 @@
 
       .section1 {
         .tokenSelect {
-          background-color: $BG02;
           .tokenItem {
             font-weight: 700;
             font-family: Inter;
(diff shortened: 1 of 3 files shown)


dkent600 (approved):

src/lbpDashboard/lbpDashboardForm/lbpDashboardForm.scss
```diff
@@ -58,16 +57,22 @@
...
               height: 63px;
               border-top-right-radius: 0;
               border-bottom-right-radius: 0;
-              background-color: transparent;
+              background-color: $BG02;
             }
 
             .menu {
               border-top-right-radius: 0;
+              padding-top: 53px;
```
dkent600, line 69: I'm wondering, are any of these fixes general to "symbol-only" and thus can be included in tokenSelect.scss instead of here, so they can potentially apply to other times we might want to use symbol-only?

dkent600 (changes requested):

src/lbpDashboard/lbpDashboardForm/lbpDashboardForm.scss
giladt: I've moved some declarations around to make the symbolOnly element more generic:

Generic Use: image  DashboardForm styling: image

dkent600 (approved):

PR #446, opened 2021-11-08
Disable funding seed that had expired

wantedsystem:


dkent600 (changes requested):
Two big problems here:

1) you've coded for LBPs, but the ticket references seeds
2) the change should be in the seed *admin* dashboard page, not the seed dashboard.

dkent600 (changes requested):
Please remove all the formatting changes that don't relate to this PR.  This is in general a best practice, thanks!

dkent600 (approved):

PR #457, opened 2021-11-08
First Data-Point is the initial launch price

giladt:


src/services/ProjectTokenHistoricalPriceService.ts
```diff
@@ -46,8 +47,10 @@ export class ProjectTokenHistoricalPriceService {
     }
 
     const startingSeconds = lbpMgr.startTime.getTime() / 1000;
-    const intervalMinutes = 60/*min*/;
-    const intervalSeconds = intervalMinutes * 60/*sec*/;
+    const timeWentLiveSeconds = (await lbpMgr.lbp.contract.getGradualWeightUpdateParams())[0].toNumber();
```
dkent600, line 50: Please get start and end times and weights from our entities.  they are all there.  Or if not, we should add them.

same thing for any contract interactions.  please keep them wrapped in the appropriate entities.

dkent600 (changes requested):

src/services/ProjectTokenHistoricalPriceService.ts
```diff
@@ -73,6 +75,15 @@ export class ProjectTokenHistoricalPriceService {
 
     const returnArray = new Array<IHistoricalPriceRecord>();
 
+    const startFundingTokenAmount = parseFloat(fromWei(lbpMgr.startingFundingTokenAmount, lbpMgr.fundingTokenInfo.decimals));
```
dkent600, line 78: Should use `this.numberService.fromString(...)` instead of parseFloat

src/services/ProjectTokenHistoricalPriceService.ts
```diff
@@ -73,6 +75,15 @@ export class ProjectTokenHistoricalPriceService {
...
+    const startProjectTokenAmount = parseFloat(fromWei(lbpMgr.startingProjectTokenAmount, lbpMgr.projectTokenInfo.decimals));
```
dkent600, line 79: `startingProjectTokenAmount` includes fees.  I suspect that is a problem here. I can change it, let me know.

dkent600 (changes requested):

dkent600 (approved):

PR #466, opened 2021-11-09
Refactor/462 make graph work

giladt:
Revert intervals to 60 mins.
Added time zone shift.
Fix to resize.
Make chart non interactive, show the whole scale.

### DIFF
--- a/src/lbpDashboard/sparkChart/spark-chart.ts
+++ b/src/lbpDashboard/sparkChart/spark-chart.ts
@@ -21,23 +21,25 @@ export class SparkChart {
   }
 
   attached(): void {
-    if (this.chart) {
-      this.chart.resize(this.container.offsetWidth, this.container.offsetHeight);
-    }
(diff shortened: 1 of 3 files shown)


src/lbpDashboard/sparkChart/spark-chart.ts
```diff
@@ -21,17 +21,14 @@ export class SparkChart {
...
+
     window.onresize = () => {
       /**
        * don't resize when the element is hidden, or height will go permanently to 0
        */
       setTimeout(() => {
-        if (this.chart) {
-          this.chart.resize(this.container.offsetWidth, this.container.offsetHeight);
-        }
+        if (this.chart) this.resizeChart();
```
dkent600, line 31: Best practice is to put everything in braces.  (Ought to be a lint rule for that).  Just for future reference....

dkent600 (approved):

PR #481, opened 2021-11-15
Add spinner when loading the chart

wantedsystem:
Result : 

https://user-images.githubusercontent.com/11537417/141811616-fe8e0200-e166-4dda-9122-cea99b59ff9b.mp4


dkent600 (changes requested):
This will not be a good solution because it will might show a spinner every time the data is updated, which is at minimum every 60 seconds.  We only want to show the spinner when entering the page.  Please follow the comments I gave in voice recordings on telegram.  If they don't make sense, ping me, thanks.

src/lbpDashboard/sparkChart/spark-chart.scss
```diff
@@ -1,3 +1,4 @@
+@import "styles/styles.scss";
```
dkent600, line 1: delete this line if no longer needed

src/lbpDashboard/lbpDashboard.html
```diff
@@ -5,6 +5,7 @@
   <require from="./sparkChart/spark-chart"></require>
   <require from="./timeRemaining/timeRemaining"></require>
   <div class="page animated-page au-animate lbpDashboardContainer">
+    <div show.to-view="loading" class="loading"><i class="fas fa-circle-notch fa-spin"></i></div>
```
dkent600, line 8: looks to me like this will affect the entire dashboard.  It should only affect the chart, and only when the chart data is being fetched on page load. See in attached:

```
 if (!lbpmgr.priceHistory) {
        this.lbpMgr.ensurePriceHistory();
      }
```

dkent600 (changes requested):

src/lbpDashboard/lbpDashboard.html
```diff
@@ -26,4 +27,4 @@
         <token-distribution start-date.bind='lbpMgr.startDate' token-details.bind='lbpMgr.metadata.tokenDetails'></token-distribution>
       </div>
     </div>
-</template>
+</template>
```
dkent600, line 30: would be nice if you could adjust you vscode settings to not remove the line feed at the end. @wantedsystem

dkent600 (approved):

dkent600:
@wantedsystem Can you please remove the merge conflicts, thanks!

PR #547, opened 2021-11-19
change subtitle text

wantedsystem:
Result =>
Capture d’écran 2021-11-19 à 10 27 45


dkent600 (changes requested):
The spec was incorrect on this, is not correct grammar.  Lets get the correct text for this from @BartuCulha

dkent600 (approved):

PR #551, opened 2021-11-19
521 Add preview chart

giladt:
Result:

preview


src/services/LbpProjectTokenPriceService.ts
```diff
@@ -1,8 +1,17 @@
+import { autoinject } from "aurelia-framework";
 import * as moment from "moment-timezone";
 
-export class LbpProjectTokenPriceService {
+enum ERoundDirection {
```
dkent600, line 4: this is overkill.  can just be a boolean variable `roundDown`, true by default

src/services/ProjectTokenHistoricalPriceService.ts
```diff
@@ -127,13 +127,11 @@ export class ProjectTokenHistoricalPriceService {
...
         }
 
-        const timezoneOffset = new Date().getTimezoneOffset() * 60;
         const priceAtTimePoint = fundingTokenPricesUSD.filter(price => price.timestamp <= timestamp );
 
-
         if (todaysSwaps?.length) {
           returnArray.push({
-            time: timestamp - timezoneOffset + intervalSeconds/* Apply to the next interval in users timezone */,
+            time: timestamp + intervalSeconds/* Apply to the next interval in users timezone */,
```
dkent600, line 134: We should document this function, and include a description such as:  "The returned data points will be relative to the timezone of the LBPManager startTime and endTime, which, according to the LBPManager entity, are in the user's timezone.

dkent600 (changes requested):

dkent600 (approved):

PR #555, opened 2021-11-21
Add Total Swap Fee to The Historical price service

giladt:
Added `lbpTotalSwapFees(poolId: string): number` method to ProjectTokenHistoricalPriceService.

Pulls the total swap fee from Balancer SubGraph, by the pool-ID.


giladt:
Please review also the fix applied to the `fetchSwap` batch iteration (first commit).

dkent600 (approved):

PR #560, opened 2021-11-22
Add max-with for cards

wantedsystem:


src/launch/launch.scss
```diff
@@ -46,6 +46,7 @@
       height: 100%;
       display: grid;
       grid-template-rows: 2fr 1fr;
+      max-width: 600px;
```
giladt, line 49: I am not sure why the cards have different widths (45%, 48%) but, you'll probably want to set the max-width on the surrounding divs rather than here, this way the cards will stay centered on wider screens.
Maybe 600px for the first div and 620px (≈+3%) for the second one?

giladt (changes requested):

src/launch/launch.scss
wantedsystem: @giladt you want me to modify this ?
Capture d’écran 2021-11-23 à 12 23 22

But the result is not good
Capture d’écran 2021-11-23 à 12 25 48

wantedsystem: @giladt but when both have max-width : 600px 
Capture d’écran 2021-11-23 à 12 27 00
it seems ok

giladt: You're right. As said, I'm not sure why there's a width difference between the cards (maybe @dkent600 remembers the reason?). I think they can both have 45% and max of 600px. Also maybe a min-width is needed to avoid this (but not on smaller screens!):

image

dkent600: I don't recall any reason for different widths.

dkent600 (approved):

dkent600 (approved):

src/launch/launch.scss
giladt: There are still some issues with the way the buttons wrap within the cards.
Under `launch-box-container` class, please apply the following CSS:
1. On `buttons` class- please add a column-gap of 10px.
2. On the `a` element decleration within the `buttons` element- please 
   a. set the left-margin to 0 and
   b. set a width of 100px.

Doing this should set the width of both buttons to be identical, align them to the left, and set a minimal gap before the `request information` button wraps.

Please test your work in all screen media queries to see if it function correctly.
Thanks!

dkent600 (approved):

PR #561, opened 2021-11-22
fix: add https in the website plaeholders

wantedsystem:
Results =>

LBP / Seed stage 1: 
Capture d’écran 2021-11-22 à 18 20 23

LBP stage 4 : 
Capture d’écran 2021-11-22 à 18 20 35

Seed stage 4 : 
Capture d’écran 2021-11-22 à 18 21 00

### DIFF
--- a/src/newLaunch/lbp/stages/stage4.html
+++ b/src/newLaunch/lbp/stages/stage4.html
@@ -197,7 +197,7 @@
           <input
             id="disclaimer"
             value.bind="launchConfig.launchDetails.legalDisclaimer"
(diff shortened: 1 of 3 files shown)


dkent600 (changes requested):
Where ever you have " : https://..." (with the colon), make it like this:

" (https://...)"

dkent600 (changes requested):
Sorry, you didn't exactly follow the instructions.  Note that the colon should be removed.

dkent600 (changes requested):
Still not right.  Missing the parens.

Don't supply parens when standalone in the placeholder. only when following other text.

PR #566, opened 2021-11-24
Add separator in wizard stage 4

wantedsystem:
Result =>

Capture d’écran 2021-11-24 à 20 29 53

### DIFF
--- a/src/newLaunch/baseStage.scss
+++ b/src/newLaunch/baseStage.scss
@@ -3,7 +3,7 @@
 // Theme for the stages
 
 .dividerFundings {
-  background-color: #3e3351;
+  background-color: $Border03;
 }
 
 .infoContainer {
--- a/src/newLaunch/baseStage.scss
+++ b/src/newLaunch/baseStage.scss
@@ -1,6 +1,11 @@
 @import "styles/styles.scss";
 // This styles are common to all the stages components
 // Theme for the stages
+
+.dividerFundings {
+  background-color: #3e3351;
+}
+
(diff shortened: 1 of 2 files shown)


src/newLaunch/baseStage.scss
```diff
@@ -1,6 +1,11 @@
 @import "styles/styles.scss";
 // This styles are common to all the stages components
 // Theme for the stages
+
+.dividerFundings {
+  background-color: #3e3351;
```
dkent600, line 6: Please always use the color variables from _colors.scss

src/newLaunch/lbp/stages/stage4.html
```diff
@@ -76,7 +76,7 @@
         <div class="infoInput maxTarget fundingTarget">
 
           <div class="fundingDescrp">
-            Lorem ipsum dolor sit amet, consectetur adipiscing elit
```
dkent600, line 79: I don't think we were supposed to remove this text.  Please check with Renc about this.

dkent600 (changes requested):

src/newLaunch/lbp/stages/stage4.html
wantedsystem: @dkent600 i asked @renc-curvelabs  and said that we need to remove it like i did

dkent600: ok

dkent600 (approved):

PR #567, opened 2021-11-24
Add space between token name and bracket

wantedsystem:
Result =>
Capture d’écran 2021-11-24 à 20 32 58
>

### DIFF
--- a/src/resources/elements/tokenSelect/tokenSelect.html
+++ b/src/resources/elements/tokenSelect/tokenSelect.html
@@ -7,7 +7,7 @@
       <div repeat.for="tokenInfo of tokenList">
         <div class="tokenItem">
           <img src.to-view="tokenInfo.logoURI"></img>
-          <div if.bind="!symbolOnly">${tokenInfo.name}</div>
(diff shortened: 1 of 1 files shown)


src/resources/elements/tokenSelect/tokenSelect.html
```diff
@@ -7,7 +7,7 @@
       <div repeat.for="tokenInfo of tokenList">
         <div class="tokenItem">
           <img src.to-view="tokenInfo.logoURI"></img>
-          <div if.bind="!symbolOnly">${tokenInfo.name}</div>
+          <div if.bind="!symbolOnly">${tokenInfo.name.split('(')[0]}&nbsp;${tokenInfo.name.split('(').slice(1).join('(')}</div>
```
dkent600, line 10: this makes no sense to me, seems way over complicated.  All we need here is a space between the name and the symbol. that should require no more than adding `&nbsp;` after the token name or before the symbol.

dkent600 (changes requested):

dkent600 (approved):

PR #595, opened 2021-11-30
Lbp dashboard should display a spinny icon when the chart is loading

wantedsystem:


dkent600:
I don't understand why this is still open.  I'm closing because it seems to me this has all been merged.  Please correct me if I'm wrong @wantedsystem

PR #633, opened 2021-12-06
610- History and forecast lines should start at the same point

giladt:
Result:

image

### DIFF
--- a/src/entities/LbpManager.ts
+++ b/src/entities/LbpManager.ts
@@ -17,7 +17,7 @@ import { ITokenInfo, TokenService } from "services/TokenService";
 import TransactionsService, { TransactionReceipt } from "services/TransactionsService";
 import { Utils } from "services/utils";
 import { Lbp } from "entities/Lbp";
-import { IHistoricalPriceRecord, ProjectTokenHistoricalPriceService, ISwapRecord } from "services/ProjectTokenHistoricalPriceService";
+import { IHistoricalPriceRecord, ProjectTokenHistoricalPriceService } from "services/ProjectTokenHistoricalPriceService";
 import { TimingService } from "services/TimingService";
 
 export interface ILbpManagerConfiguration {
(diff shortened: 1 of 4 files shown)


src/entities/LbpManager.ts
```diff
@@ -488,6 +489,7 @@ export class LbpManager implements ILaunch {
           .then((history) => {
             // this.ensureTrajectoryForecastData(history);
             this.priceHistory = history;
+            this.lastSwap = this.projectTokenHistoricalPriceService.lastSwap;
```
dkent600, line 492: why must this be here?  is it not easily accessible in the code where it is actually needed?

dkent600 (changes requested):

src/entities/LbpManager.ts
giladt: The forecast data is being changed every time I attach the chart.
I need to ensure that the last swap I pass to it is the actual LBP last swap, otherwise, it uses the value initialized in the first initialization.

dkent600 (changes requested):
Sending this back per our phone convo

src/entities/LbpManager.ts
giladt: removed from entity

src/lbpDashboard/lbpPriceChart/lbpPriceChart.ts
```diff
@@ -35,20 +35,20 @@ export class LbpPriceChart {
...
+    if ( this.lbpMgr ) {
+      this.dataPropertyWatcher = this.aureliaHelperService.createPropertyWatch(this.lbpMgr, "priceHistory", this.hydrateChart);
+    }
   }
 
   private graphConfig: Array<any>;
 
   private async hydrateChart(reset = false): Promise<void> {
-    if (this.lbpMgr) {
+    if (this?.lbpMgr) {
```
dkent600, line 51: how can this '?' be needed?

src/services/ProjectTokenHistoricalPriceService.ts
```diff
@@ -12,10 +12,11 @@ import { jsonToGraphQLQuery } from "json-to-graphql-query";
 import axios from "axios";
 import { autoinject } from "aurelia-framework";
 
-interface ISwapRecord {
+export interface ISwapRecord {
```
dkent600, line 15: export still needed?

dkent600 (changes requested):

src/services/ProjectTokenHistoricalPriceService.ts
giladt: removed

PR #658, opened 2021-12-10
653 - Display the preview chart on the lbp dashboard until launch begins

giladt:
Fixed the chart issue- shows now the trajectory data for launches that haven't started yet.

Added fix to the timescale formatting with the following logic:
Under 24h duration: "H:mm" (5:00)
Under 5 days duration: "MMM DD | H:mm" (Dec 10 | 17:00)
Above 5 days duration: "MMM DD" (Dec 10)

Added Border to the timescale at bottom of the chart (Dashboard only!)
Removed the border from the price scale at the right - Previously, it was in an undefined color. (Coordinated with @renc-curvelabs )

Result: (Mainnet | D2D launch)
image

image

image

### DIFF
--- a/src/services/ProjectTokenHistoricalPriceService.ts
+++ b/src/services/ProjectTokenHistoricalPriceService.ts
@@ -222,7 +222,7 @@ export class ProjectTokenHistoricalPriceService {
      * calculated forecast to the end of the array up to the current time.
      */
     const forecastData = await this.getTrajectoryForecastData(lbpMgr);
-    const pastForecast = forecastData?.filter((item) => item.time < (currentTime));
+    const pastForecast = forecastData?.filter((item) => item.time < currentTime);
     returnArray.pop(); // avoid duplicate last item
     return [...returnArray, ...pastForecast];
   }
@@ -243,7 +243,7 @@ export class ProjectTokenHistoricalPriceService {
     const prices = await this.getFundingTokenUSDPricesByID(
       lbpMgr.fundingTokenInfo.id,
       endTimeSeconds,
-      startTimeSeconds < currentTime ? startTimeSeconds : Math.floor(currentTime) - 3600,
+      (startTimeSeconds < currentTime) ? startTimeSeconds : (Math.floor(currentTime) - 3600),
       intervalMinutes,
     );
 
--- a/src/services/ProjectTokenHistoricalPriceService.ts
+++ b/src/services/ProjectTokenHistoricalPriceService.ts
@@ -73,7 +73,7 @@ export class ProjectTokenHistoricalPriceService {
    * @returns Array(IHistoricalPriceRecord): {time: number, price?: number}
    */
   public async getPricesHistory(lbpMgr: LbpManager): Promise<Array<IHistoricalPriceRecord>> {
-    if (!lbpMgr.lbp || !lbpMgr.lbp.poolId || lbpMgr.startTime.getTime() >= new Date().getTime()) {
+    if (!lbpMgr.lbp || !lbpMgr.lbp.poolId || lbpMgr.hasNotStarted) {
       return [];
     }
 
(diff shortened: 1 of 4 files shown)


src/services/ProjectTokenHistoricalPriceService.ts
```diff
@@ -73,7 +73,7 @@ export class ProjectTokenHistoricalPriceService {
    * @returns Array(IHistoricalPriceRecord): {time: number, price?: number}
    */
   public async getPricesHistory(lbpMgr: LbpManager): Promise<Array<IHistoricalPriceRecord>> {
-    if (!lbpMgr.lbp || !lbpMgr.lbp.poolId) {
+    if (!lbpMgr.lbp || !lbpMgr.lbp.poolId || lbpMgr.startTime.getTime() >= new Date().getTime()) {
```
dkent600, line 76: use `!lbpMgr.hasNotStarted`

src/services/ProjectTokenHistoricalPriceService.ts
```diff
@@ -113,25 +113,25 @@ export class ProjectTokenHistoricalPriceService {
     const prices = await this.getFundingTokenUSDPricesByID(
       lbpMgr.fundingTokenInfo.id,
       endTimeSeconds,
-      startTimeSeconds,
+      startTimeSeconds < new Date().getTime() / 1000 ? startTimeSeconds : Math.floor(new Date().getTime() / 1000) - 3600,
```
dkent600, line 116: Generally is most efficient and most consistent in having a single sense of "now", to define `now` once at the top of a function instead of calling new Date over and over again, inefficiently and potentially giving you different values.

dkent600 (changes requested):

src/services/ProjectTokenHistoricalPriceService.ts
```diff
@@ -73,7 +73,7 @@ export class ProjectTokenHistoricalPriceService {
    * @returns Array(IHistoricalPriceRecord): {time: number, price?: number}
    */
   public async getPricesHistory(lbpMgr: LbpManager): Promise<Array<IHistoricalPriceRecord>> {
-    if (!lbpMgr.lbp || !lbpMgr.lbp.poolId || lbpMgr.startTime.getTime() >= new Date().getTime()) {
+    if (!lbpMgr.lbp || !lbpMgr.lbp.poolId || lbpMgr.hasNotStarted) {
```
dkent600, line 76: Is supposed to be `!lbpMgr.hasNotStarted`  (see the '!')

dkent600 (changes requested):

src/services/ProjectTokenHistoricalPriceService.ts
```diff
@@ -222,7 +222,7 @@ export class ProjectTokenHistoricalPriceService {
      * calculated forecast to the end of the array up to the current time.
      */
     const forecastData = await this.getTrajectoryForecastData(lbpMgr);
-    const pastForecast = forecastData?.filter((item) => item.time < (new Date().getTime() / 1000));
+    const pastForecast = forecastData?.filter((item) => item.time < (currentTime));
```
dkent600, line 225: as long as we're making changes, `(currentTime)` would look a little less silly without the parens.

src/services/ProjectTokenHistoricalPriceService.ts
```diff
@@ -243,7 +243,7 @@ export class ProjectTokenHistoricalPriceService {
     const prices = await this.getFundingTokenUSDPricesByID(
       lbpMgr.fundingTokenInfo.id,
       endTimeSeconds,
-      startTimeSeconds,
+      startTimeSeconds < currentTime ? startTimeSeconds : Math.floor(currentTime) - 3600,
```
dkent600, line 246: Maybe is fine as is, but my paranoia always makes me use parens around the expressions in terniary expressions:

`(startTimeSeconds < currentTime) ? startTimeSeconds : (Math.floor(currentTime) - 3600)`

src/services/ProjectTokenHistoricalPriceService.ts
giladt: hasNotStarted returns `true` 













;

dkent600: my brain is just too ready for a break

dkent600 (approved):

PR #669, opened 2021-12-14
Chore : fixed the backling issue, now it points to lbp instead of lbpmanager

codebuster22:


src/launches/launches.ts
```diff
@@ -90,11 +91,22 @@ export class Launches {
   }
 
   gotoEtherscan(launch: ILaunch, event: Event): boolean {
-    Utils.goto(this.ethereumService.getEtherscanLink(launch.address));
+    if (launch.launchType === LaunchType.LBP ){
+      Utils.goto(this.ethereumService.getEtherscanLink(this.getLbpAddress(launch.address)));
```
dkent600, line 95: should just be:  `Utils.goto(this.ethereumService.getEtherscanLink((launch as LbpManager).lbp.address));

(or some type cast like that)

src/launches/launches.ts
```diff
@@ -90,11 +91,22 @@ export class Launches {
...
+    } else {
+      Utils.goto(this.ethereumService.getEtherscanLink(launch.address));
+    }
     event.stopPropagation();
     return false;
   }
 
+  getLbpAddress(lbpManager: Address): Address{
```
dkent600, line 103: remove

src/launches/launches.ts
```diff
@@ -1,5 +1,5 @@
 import { EventAggregator } from "aurelia-event-aggregator";
-import { EthereumService } from "services/EthereumService";
+import { EthereumService, Address } from "services/EthereumService";
```
dkent600, line 2: can remove Address?

dkent600 (changes requested):

src/launches/launches.ts
codebuster22: removed

src/launches/launches.ts
codebuster22: removed

src/launches/launches.ts
codebuster22: implemented your suggestion

dkent600 (approved):

PR #686, opened 2021-12-15
Add link to documentation

rogervs:


dkent600:
Thanks much for the contribution, @rogervs !

The link you supplied explains how to contribute to the $D2D token raise. The content is not relevant at the moment as the token launch has ended. However, there is good content here that we can re-use as a general how-to-guide to contribute to LBPs 

It needs retouching and should be treated as a separate documentation task. We should not include this link in the documentation page now.

PR #687, opened 2021-12-16
685 - Include swap fee in the chart price calculations

giladt:


src/services/ProjectTokenHistoricalPriceService.ts
```diff
@@ -237,17 +237,19 @@ export class ProjectTokenHistoricalPriceService {
     const poolPTBalance = this.numberService.fromString(fromWei(lbpMgr.lbp.vault.projectTokenBalance, lbpMgr.projectTokenInfo.decimals));
     const poolFTBalance = this.numberService.fromString(fromWei(lbpMgr.lbp.vault.fundingTokenBalance, lbpMgr.fundingTokenInfo.decimals));
     const projectTokenAmount = this.numberService.fromString(lastSwap.tokenAmountOut);
-    // const fundingTokenAmount = this.numberService.fromString(lastSwap.tokenAmountIn);
     const forecastData = await this.priceService.getInterpolatedPriceDataPoints(
       poolPTBalance,
-      poolFTBalance,
+      poolFTBalance * (1 + lbpMgr.swapFeePercentage),
```
fabianschu, line 242: I don't think this is necessary. When we get the `poolFTBalance` from the vault, my understanding is that all swap fees that had been collected so far are included in that `poolFTBalance`.

src/services/ProjectTokenHistoricalPriceService.ts
```diff
@@ -159,31 +159,31 @@ export class ProjectTokenHistoricalPriceService {
...
+            (this.numberService.fromString(todaysSwaps[todaysSwaps.length-1].tokenAmountIn) * (1 + lbpMgr.swapFeePercentage)) /
             (this.numberService.fromString(todaysSwaps[todaysSwaps.length-1].tokenAmountOut)) *
-            priceAtTimePoint
+            usdPriceAtTimePoint
           ),
         });
 
         previousTimePoint = (
-          (this.numberService.fromString(todaysSwaps[todaysSwaps.length-1].tokenAmountIn)) /
+          (this.numberService.fromString(todaysSwaps[todaysSwaps.length-1].tokenAmountIn) * (1 + lbpMgr.swapFeePercentage)) /
```
fabianschu, line 173: Just to make sure: the swap fee is not already included in the swap data that we receive from the subgraph per swap?

dkent600: correct, we confirmed this looking at the contracts.  wouldn't hurt if you want to triple check @fabianschu .

src/services/ProjectTokenHistoricalPriceService.ts
dkent600: You are correct @fabianschu

giladt: removed the fee from pool balance

dkent600 (approved):

fabianschu (commented):
looks good to me

PR #698, opened 2021-12-21
688 - Time left display (fix and feature change)

hiaux0:
_(quick introduction: Hello curious reader, I'll be joining Curvelabs officially on the 1st Jan 2022 as a Frontend Engineer)_

## What was done
a. fix #688 by adding `largest` to `projectTokenInfo.html` e6f6cd96558d1534b0f1732f766e643e102962c0
b. change how `largest` behaves a71d3df0885930202dae20b0209f01348d73de4e

## Testing
http://localhost:3330/lbp/0x7c10deE4389feE638e2E84dC1671B25d7DF2935f (testing url)
1. Navigate to `/launches` route
2. select a type `lpb`

#### Before
Only showed hours
image

#### After
Show hours and minutes (aka. the largest 2 
image

#### Before
image

#### After
Only 2 largest
image

/launches
image

##### Responsiveness check

https://user-images.githubusercontent.com/30693990/146975648-9c63770f-f427-492e-bf04-8153e9447a25.mp4

Issue #688, opened 2021-12-17
Time left should show how many minutes left


BartuCulha:

dkent600:
And seconds.  Should show the largest unit at the current point in time.

dkent600:
@BartuCulha How much time was remaining when you saw this?

BartuCulha:
@dkent600 its shown in the screenshot. Time remaining was 1 hours 31 minutes. Best way to reproduce is to create a new LBP that lasts less than 24 hours

PR #701, opened 2022-01-05
Add `largest2` to DateService

hiaux0:
Relates to #698 

## What was done
- add `largest2` flag to DateService
- added tests
- (difference to #698 is, that here, we don't change the UI)

@dkent600 
I decided to create a new PR, that only introduces the `largest2` flag.
We keep the `largest` behavior.
As for the original issue #688, imo we should have a final call for that with @BartuCulha


test/unit/services/DateService.spec.ts
```diff
@@ -0,0 +1,149 @@
...
+      [TimespanResolution.largest2, [1, 0, 1, 1], "1 day, 1 minute"],
+      [TimespanResolution.largest2, [1, 0, 1, 0], "1 day, 1 minute"],
+      [TimespanResolution.largest2, [1, 0, 0, 1], "1 day, 1 second"],
+      [TimespanResolution.largest2, [1, 0, 0, 0], "1 day, 0 seconds"],
+      [TimespanResolution.largest2, [0, 1, 1, 1], "1 hour, 1 minute"],
+      [TimespanResolution.largest2, [0, 1, 1, 0], "1 hour, 1 minute"],
+      [TimespanResolution.largest2, [0, 1, 0, 1], "1 hour, 1 second"],
+      [TimespanResolution.largest2, [0, 1, 0, 0], "1 hour, 0 seconds"],
+      [TimespanResolution.largest2, [0, 0, 1, 1], "1 minute, 1 second"],
+      [TimespanResolution.largest2, [0, 0, 1, 0], "1 minute, 0 seconds"],
```
hiaux0, line 38: @dkent600 
We talked about it a bit, but here more formalized:
What should be the desired output for this case? (where amount of milliseconds is exactly 1 minute)

Imo, to only have minimal changes, we can omit this corner case

dkent600 (approved):

PR #773, opened 2022-09-07
Feature/768 workaround for celo

testing-life:
for celo users, the result should be:
image

for others, the result should be 
image

also, fixed token symbol not appearing in thank you page
image


src/newLaunch/stages/stage8.html
```diff
@@ -1,63 +1,91 @@
...
 
           will have to fund the ${wizardState.launchTypeTitle} with
 
           <span if.bind="wizardState.requiredProjectTokenDeposit">
             <formatted-number
               value.to-view="wizardState.requiredProjectTokenDeposit  | ethwei:launchConfig.tokenDetails.projectTokenInfo.decimals"
-              thousands-separated></formatted-number>
+              thousands-separated
+            ></formatted-number>
             ${launchConfig.tokenDetails.projectTokenInfo.symbol}
```
giladt, line 56: The symbol is not shown. Can you check that?
image

src/newLaunch/stages/stage8.ts
```diff
@@ -24,4 +27,10 @@ export class Stage8 extends BaseStage<ILaunchConfig> {
   get ipfsURL(): string {
     return this.ipfsService.getIpfsUrl(this.wizardState.launchHash);
   }
+
+  attached():void{
+    if (EthereumService.targetedNetwork === Networks.Celo || EthereumService.targetedNetwork === Networks.Alfajores) {
```
giladt, line 32: Can use `isCeloLikeNetwork`?

giladt (changes requested):

src/newLaunch/stages/stage8.html
testing-life: hm so with D2D it's correct. i'll check rinkeby with GILAD
image

src/newLaunch/stages/stage8.html
```diff
@@ -1,63 +1,91 @@
...
-          appear in and be used in the Prime Launch interface. The approval takes place in the <a
-            href.bind="multiSigWalletUri" target="_blank" rel="noopener noreferrer">Prime Launch multisig wallet</a>.
-          Your ${wizardState.launchTypeTitle} may take up to 5 minutes to appear in the multisig. Reviews generally take
-          5-7 days."</p>
+        <p if.to-view="dataToken">
+          <span class="importantText"
+            >Note that when you generate a new token, the value will only be displayed once, so be sure to write it
+            down</span
+          >
+          <span class="dataTokenWrapper">${dataToken}</span>
```
hiaux0, line 21: I left a comment about html wrap formatter from vscode.
Can you check if that's the case for you?

Ideally, not major formatting should happen

hiaux0 (approved):
the html formatting really makes it hard to spot the actual changes

src/newLaunch/stages/stage8.html
testing-life: that's the correct setting, no ?
image

giladt (dismissed):

src/newLaunch/stages/stage8.html
hiaux0: yeah, then I don't know why you have all those html attribute changes?

hiaux0 (dismissed):

PR #783, opened 2022-09-08
Celo-transaction test

giladt:
Test to try transactions in Celo
Tested and works for both Alfajores and Celo networks.

navigate to `/celo-transation`

### DIFF
--- a/src/test/celoTransaction/celoTransaction.ts
+++ b/src/test/celoTransaction/celoTransaction.ts
@@ -1,0 +1,83 @@
+import { autoinject, computedFrom } from "aurelia-framework";
+import { EventAggregator } from "aurelia-event-aggregator";
+import { BigNumber, ethers } from "ethers";
+import { EthereumService } from "services/EthereumService";
+import { DisposableCollection } from "services/DisposableCollection";
+import { TokenService } from "services/TokenService";
+import TransactionsService from "services/TransactionsService";
+import "./celoTransaction.scss";
+
+@autoinject
+export class celoTransaction {
+  private subscriptions = new DisposableCollection();
+
+  constructor(
+    private eventAggregator: EventAggregator,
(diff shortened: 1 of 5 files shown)


hiaux0:
Thx for verifying!
Moved the underlying ticket to Code Complete

src/test/celoTransaction/celoTransaction.ts
```diff
@@ -0,0 +1,83 @@
...
+      try {
+        this.checking = true;
+        if (this.account) {
+          const provider = this.ethereumService.readOnlyProvider;
+          this.balance = await provider.getBalance(this.account);
+        } else {
+          this.balance = null;
+        }
+        // tslint:disable-next-line:no-empty
+        // eslint-disable-next-line no-empty
```
testing-life, line 46: dont suppose  we could jsut use one of those, no ?

testing-life: also, wouldn't we want to at least log an error then it happens here ? otherwise,the try catch block kinda is pointless, i feel

src/test/celoTransaction/celoTransaction.ts
```diff
@@ -0,0 +1,83 @@
...
+      }));
+  }
+
+  handleMaxBuy() {
+    this.tokenAmountToPay = this.balance ?? BigNumber.from(0);
+  }
+
+  async doTransaction() {
+    const provider = this.ethereumService.walletProvider;
+    const gasPrice = provider.getGasPrice();
```
testing-life, line 68: are we sure all those function wont fail, for whatever reason? i wonder if it would be beneficial to try catch those here. or structure it a bit more defensively

giladt:
Converted this PR to draft, as it is meant for research purposes and is not intended to be merged.

PR #797, opened 2022-09-14
Sc 300/add csv upload input to seed details stage

testing-life:
image

needs to be revisited after BE adds aan allow list param
```
const csvContent = this.csv && await this.csv[0].text();
    // for BE adds allow list param
    // this.launchConfig.launchDetails.allowList = csvContent;
```

with fake label
image

### DIFF
--- a/src/newLaunch/baseStage.scss
+++ b/src/newLaunch/baseStage.scss
@@ -461,6 +461,15 @@
       align-items: center;
       flex-wrap: wrap;
     }
+    
+    .csvUploadInputContainer {
+      @extend .adminInputContainer;
+      gap: 15px;
+
+      input[id=allowList]::file-selector-button {
+        display: none;
+      }
+    }  
     .button2 {
       margin-left: 0;
     }
--- a/src/newLaunch/baseStage.scss
+++ b/src/newLaunch/baseStage.scss
@@ -37,11 +37,9 @@
   .content-block {
     grid-template-columns: 1fr;
   }
-
   .footer-input {
     max-width: 50%;
   }
-
   #footer-date-content {
     display: grid;
     grid-template-columns: 1fr;
@@ -85,12 +83,13 @@
   text-decoration: none;
 }
 
-@mixin boldButton {
+.make-admin-button {
   display: inline-block;
   border-radius: 10px;
   border: none;
   outline: none;
   padding: 15px;
+  margin-top: 40px;
   font-family: Inter;
   font-style: normal;
   font-weight: 500;
@@ -102,16 +101,7 @@
   background-color: rgba(64, 52, 83, 1);
   color: $White;
   cursor: pointer;
-}
-
-.make-admin-button {
-  @include boldButton;
-  margin-top: 40px;
   margin-bottom: 80px;
-}
-
-.upload-csv {
-  @include boldButton;
 }
 
 .end-stage-subtext {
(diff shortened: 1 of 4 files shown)


src/newLaunch/baseStage.scss
```diff
@@ -37,9 +37,11 @@
   .content-block {
```
hiaux0, line 37: Can you please revert all unrelated formatting changes?

testing-life: done

src/newLaunch/seed/stages/stage4.html
```diff
@@ -120,6 +120,22 @@
...
+                type="file"
+                accept=".csv"
+                files.bind="csv"
+                placeholder="Upload your .csv allowlist..." />
+              </div>
+            </span>
+              <div >
+                <button class="upload-csv" click.delegate="parseCsv()">Upload</button>
+                </div>
+          </div>
```
hiaux0, line 138: formatting off for this whole section, can you adjust please?

src/newLaunch/seed/stages/stage4.html
```diff
@@ -120,6 +120,22 @@
...
+                <input class="footer-input"
+                id="allowList"
+                type="file"
+                accept=".csv"
+                files.bind="csv"
+                placeholder="Upload your .csv allowlist..." />
+              </div>
+            </span>
+              <div >
+                <button class="upload-csv" click.delegate="parseCsv()">Upload</button>
```
hiaux0, line 136: I expected to have have the file upload popup appear when pressing on "Upload".
(currently popup appears when clicking on the input)

Can we have an UX like this
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file

I added 2 Tasks to the ticket
https://app.shortcut.com/curvelabs/story/300/add-csv-upload-input-to-seed-details-stage-4

hiaux0 (commented):
## Fnality
- Should be on 2 lines (label and input)
image

src/newLaunch/baseStage.scss
```diff
@@ -101,9 +134,18 @@
   background-color: rgba(64, 52, 83, 1);
   color: $White;
   cursor: pointer;
+}
+
+.make-admin-button {
```
giladt, line 139: I suggest we follow only one naming convention. I guess camelCase is should be used across `prime-launch`.

giladt (changes requested):

src/newLaunch/baseStage.scss
testing-life: there seems to be a mix of naming conventions used - kebab case, some bem naming, camel case. this change was introduced in july
image

this specific class `.make-admin-button` was also introduced by someone else, also in july. my point is, your request would entail a refactor, which we do not want to do right now, correct ?

src/newLaunch/seed/stages/stage4.html
testing-life: how about we move and do as giald suggests: 
_At this point probably best to not having any formatter, IMO… or once dev is merged into 756, we can create a PR that format’s all files to specific config and continue with unified EditorConfig from there?_
I don't know what settings to use to format it to your expectations

src/newLaunch/seed/stages/stage4.html
testing-life: same mdn article displays this pattern, which is closer to our behaviour (minus the browser button):
image

src/newLaunch/baseStage.scss
giladt: Make sense. Probably should be done together (or alongside) the formatting task you've suggested in the other comment...

testing-life: :+1:

hiaux0 (approved):

PR #802, opened 2022-09-16
Sc 351/wizard summary update classes feature from

testing-life:
update based on figma design
some values are hardcoded - need to be revisited
image


hiaux0 (changes requested):
## Code
- Can you add TODO to the hardcoded places?

## Fnality
- Start Date and Start Time distance (end as well)
  - mocks have 60px distance 
image

PR #803, opened 2022-09-20
575 - refac(wizard.seed.thankyou): remove tx data field (used for celo)

hiaux0:
## What was done
- remove tx data field from Thank you page in Wizard>Seed

## Testing
- create seed
- reach thank you page

### Before
- showed data field

### After
- data field gone
image

### DIFF
--- a/src/newLaunch/baseStage.scss
+++ b/src/newLaunch/baseStage.scss
@@ -140,15 +140,6 @@
   .importantText {
     font-weight: 800;
   }
-
-  .dataTokenWrapper {
-    word-wrap: anywhere;
-    display: block;
-    padding: 15px;
-    background: rgba(0, 0, 0, 0.5);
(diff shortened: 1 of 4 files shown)


src/newLaunch/baseStage.scss
```diff
@@ -140,15 +140,6 @@
   .importantText {
```
hiaux0, line 140: (just general comment)
@giladt 
Can you cross-check with PR #773 
there were some unrelated changes, but they seem to be good to keep

giladt (approved):
Looks good to me.
The only thing to consider: The stages validation has been previously done in stage7, but since we needed to pass the IPFS hash after the submission, It's being done there, upon detach.
Though I believe it's cleaner and better this way, it also might make sense to restore the original code, if this change is not truly needed. WDYT?

PR #804, opened 2022-09-20
570, 572 - summary page token address `undefined` and open in new tab

hiaux0:
[sc-570]
[sc-572]

fix bugs as described in ticket


giladt (approved):
Looks good.

PR #810, opened 2022-09-23
dev(console): add method to easily log messages

hiaux0:


giladt (approved):
> TODO: maybe have to care for, that not too many listeners will be added
I think we use the `subscriptions` disposable collection for that, right?

PR #811, opened 2022-09-23
fix(wizard.summary): funding decimals

hiaux0:
## What was done
@giladt Wdyt about this PR, it was wrong before, was it?

### Before
- Stage Seed Details
image

- Summary
image

### After
image


giladt (approved):
Seems like it was wrong. Should be FundingToken.

PR #821, opened 2022-09-28
[Admin Dashboard | feature/sc-999]: admin-dashboard-implement-deploy-whitelist

giladt:
## What was done
Added deployment of the `allowed list` to the contract.

### How it works:
- TX1- The new classes are / the edited class is being deployed.
if we get a receipt:
- TX2- the collected addresses for the class are being deployed.


giladt:
Closed due to contract change

PR #828, opened 2022-10-07
Bug/sc 1012/modal adjust pixel perfect to mocks or at

giladt:
## What was done
* Fixed gradient border (NOTE: This change applies app-wide for all the modals! It fixes an issue with the border leaking around the modal body.
* Fixed font size/color/weight
* Fixed `line height` for labels
* Fixed fields and modal width
* Added Cancel button
* Remove Delete button (on edit) - not in the mocks
* Applied new copy for labels and tooltips
* Fixed Class Name max length to 32 (As commented in ticket SC-739)

### Result
image

### DIFF
--- a/src/resources/dialogs/addClass/addClass.html
+++ b/src/resources/dialogs/addClass/addClass.html
@@ -20,13 +20,13 @@
                 <question-mark text="TODO: Add text"></question-mark>
               </div> -->
               <span>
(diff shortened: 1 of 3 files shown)


hiaux0 (approved):
looks good.

will not merge as you self-assigned

